PR #60, opened 2014-12-17
Implement Subscription/Subscription Update Preview API calls

cgerrior:
The RecurlyClient class is missing the ability to [preview a new subscription](https://docs.recurly.com/api/subscriptions#preview-sub) or to [preview a change to an existing subscription] (https://docs.recurly.com/api/subscriptions#preview-sub). Since these API calls return the same type of response as the [createSubscription](https://github.com/cgerrior/recurly-java-library/blob/master/src/main/java/com/ning/billing/recurly/RecurlyClient.java#L248) and [updateSubscription](https://github.com/cgerrior/recurly-java-library/blob/master/src/main/java/com/ning/billing/recurly/RecurlyClient.java#L324) methods do currently, I essentially copied those and modified the URL that is called appropriately.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -288,6 +288,7 @@
                     <!--
                       | Apple's JVM sometimes requires more memory
                     -->
+		    <additionalparam>-Xdoclint:none</additionalparam>
                     <additionalJOption>-J-Xmx1024m</additionalJOption>
                     <source>1.6</source>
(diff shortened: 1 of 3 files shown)


pierre:
Fixed by 11638ea4605f40e9e7c04f3927a6b87cc3961cee.

PR #61, opened 2014-12-31
Mark invoice as paid

cgerrior:
Added the ability to post an invoice to an account, allowing you to invoiced pending adjustments. Also added the ability to mark an invoice as collected or failed.


pierre:
Fixed by e7645f173b506a41cee70f5c71f8b14bedc75a99.

PR #62, opened 2014-12-31
Collection method sub update

cgerrior:
Added the "collection_method" attribute to the SubscriptionUpdate class.

### DIFF
--- a/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
+++ b/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
@@ -701,6 +701,7 @@ public class TestRecurlyClient {
             final SubscriptionUpdate subscriptionUpdateData = new SubscriptionUpdate();
             subscriptionUpdateData.setTimeframe(SubscriptionUpdate.Timeframe.now);
             subscriptionUpdateData.setPlanCode(plan2.getPlanCode());
+			subscriptionUpdateData.setCollectionMethod(SubscriptionUpdate.CollectionMethod.manual);
 			
 			// Preview the subscription update
 	        final Subscription subscriptionUpdatedPreview = recurlyClient.updateSubscriptionPreview(subscription.getUuid(), subscriptionUpdateData);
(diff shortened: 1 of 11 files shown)


src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -681,18 +683,23 @@ public void testUpdateSubscriptions() throws Exception {
...
+			subscriptionUpdateData.setCollectionMethod("manual");
+						
+			//Update the subscription
             final Subscription subscriptionUpdated = recurlyClient.updateSubscription(subscription.getUuid(), subscriptionUpdateData);
 
             Assert.assertNotNull(subscriptionUpdated);
             Assert.assertEquals(subscription.getUuid(), subscriptionUpdated.getUuid());
             Assert.assertNotEquals(subscription.getPlan(), subscriptionUpdated.getPlan());
             Assert.assertEquals(plan2.getPlanCode(), subscriptionUpdated.getPlan().getPlanCode());
+			Assert.assertEquals("manual",subscriptionUpdated.getCollectionMethod());
```
pierre: Does this test pass for you?

On my test tenant, the API doesn't return a `collection_method` field in the _Update subscription_ response. Also, looking at the docs (https://docs.recurly.com/api/subscriptions#update-subscription), it seems that the field is only returned in the _Preview a subscription change_ response.

cgerrior: Yes, it does pass. The API docs are incomplete. I’ll suggest to my colleagues that this section be updated. The “collection_method” attribute is returned as part of the update subscription call. See the attached image for confirmation.



Thanks!

Charlie Gerrior
Sales Engineer | Recurly

charlie@recurly.com
P (415) 683-7075
recurly.com

pierre: Weird, I cannot make it pass on my tenant (the `collection_method ` attribute is not specified in the response). I've pushed the model changes regardless but commented the test.

pierre:
Fixed by 47dd716184259ff47746d50853f689a8dd54bd41.

cgerrior:
@pierre Setting a subscription (or invoice) to manual collection will only work on Recurly sites for which the corresponding feature is enabled. My test site has the feature enabled, but it's typically only available for Recurly customers who subscribe to our "Enterprise" level plan. I'm not sure how you wish to handle a scenario like this.

pierre:
That makes sense, thanks for the clarification.

I think having an implementation of the Enterprise features in this library would be valuable. We can document which ones are Enterprise-only in the Javadocs and we should probably have a special `TestRecurlyClientEnterprise` series of regression tests.

PR #64, opened 2015-01-12
update get account and get invoice methods on subscription object

cgerrior:
Method to get invoice from subscription object for both nested and linked invoices.


pierre:
Thanks!

PR #65, opened 2015-01-30
adds the ability to apply coupon at subscription creation

cgerrior:
Adds the functionality of [applying a coupon to an account when a subscription is created] (https://docs.recurly.com/api/subscriptions#create-subscription). Tests included.


pierre:
Thanks!

PR #67, opened 2015-02-03
adds the ability to enter offline payments for manual invoices

cgerrior:
(enterprise feature). Tests included. https://docs.recurly.com/api/invoices#offline-manual-payment.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -542,7 +542,7 @@ public class RecurlyClient {
     }
 
     /**
-     * Mark an invoice as paid successfully
+     * Mark an invoice as paid successfully - Recurly Enterprise Feature
      *
      * @param invoiceId Recurly Invoice ID
      */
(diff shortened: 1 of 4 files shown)


src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -587,6 +587,37 @@ public void testCreateAndCloseInvoices() throws Exception {
 
             final Invoice failedInvoice = recurlyClient.markInvoiceFailed(invoiceFail.getInvoiceNumber());
             Assert.assertEquals(failedInvoice.getState(), "failed", "Invoice not failed successfully");
+			
+            // Create an Adjustment
+            final Adjustment c = new Adjustment();
```
pierre: Could you put the enterprise-only checks in their own tests and group, so I can still run the `integration` ones before releases?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -558,6 +558,18 @@ public Invoice markInvoiceSuccessful(final Integer invoiceId) {
     public Invoice markInvoiceFailed(final Integer invoiceId) {
         return doPUT(Invoices.INVOICES_RESOURCE + "/" + invoiceId + "/mark_failed", null, Invoice.class);
     }
+	
+	/**
+	* Enter an offline payment for a manual invoice (beta)
```
pierre: Let's add a comment in the Javadoc that it's an enterprise-only feature.

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
cgerrior: I think I've done it correctly. I'll make sure to add any enterprise features to the correct test in the future. Thanks!

pierre: Perfect! The only modification I've done was to remove the `alwaysRun = true` from the `@BeforeMethod` and `@AfterMethod` annotations, and list the groups explicitly (that way, they are not run when running the `fast` tests).

FYI, interestingly enough, the test `testCreateAndCloseManualInvoices` now passes on my test tenant.

PR #68, opened 2015-03-16
adds sub notes

cgerrior:
Adds the ability to define subscription notes (for purposes of customer communication) for a new subscription. Also allows you to [update an existing subscription's notes](https://docs.recurly.com/api/subscriptions#update-subscription-notes).


pierre:
Thanks!

PR #69, opened 2015-04-13
Replaces "api.recurly.com" w/ ":subdomain.recurly.com" for easier troubleshooting.

cgerrior:
Recurly API requests should be made to ":subdomain.recurly.com" rather than "api.recurly.com". This allows for easier troubleshooting.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -127,9 +127,13 @@ public class RecurlyClient {
     private final String key;
     private final String baseUrl;
     private AsyncHttpClient client;
-
+    
     public RecurlyClient(final String apiKey) {
-        this(apiKey, "api.recurly.com", 443, "v2");
+        this(apiKey, "api");
+    }
+
+    public RecurlyClient(final String apiKey, final String subDomain) {
+        this(apiKey, subDomain + ".recurly.com", 443, "v2");
     }
 
     public RecurlyClient(final String apiKey, final String host, final int port, final String version) {
--- a/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
+++ b/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
@@ -56,6 +56,7 @@ public class TestRecurlyClient {
 
     public static final String RECURLY_PAGE_SIZE = "recurly.page.size";
     public static final String KILLBILL_PAYMENT_RECURLY_API_KEY = "killbill.payment.recurly.apiKey";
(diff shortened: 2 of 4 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -128,8 +128,8 @@ public static String getPageSizeGetParam() {
     private final String baseUrl;
     private AsyncHttpClient client;
 
-    public RecurlyClient(final String apiKey) {
-        this(apiKey, "api.recurly.com", 443, "v2");
+    public RecurlyClient(final String apiKey, final String subDomain) {
```
pierre: Would it make sense to keep the old constructor for backward compatibility?

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -67,12 +68,16 @@
     @BeforeMethod(groups = {"integration", "enterprise"})
     public void setUp() throws Exception {
         final String apiKey = System.getProperty(KILLBILL_PAYMENT_RECURLY_API_KEY);
+        final String subDomain = System.getProperty(KILLBILL_PAYMENT_RECURLY_SUBDOMAIN);
         if (apiKey == null) {
             Assert.fail("You need to set your Recurly api key to run integration tests:" +
                         " -Dkillbill.payment.recurly.apiKey=...");
+        } if (subDomain == null) {
```
pierre: Maybe for testing we could simply default to `api`, to simplify the setup?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
cgerrior: It probably does :). I'm not sure of the best way to handle that, however. What do you suggest?

pierre: The classic way is to provide two constructors:

```
    public RecurlyClient(final String apiKey) {
        this(apiKey, "api");
    }

    public RecurlyClient(final String apiKey, final String subDomain) {
        this(apiKey, subDomain + ".recurly.com", 443, "v2");
    }
```

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
cgerrior: Sure. Take a peek now.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
cgerrior: I've implemented this exactly as you've suggested. Thanks for the tip!

pierre:
:beers:

cgerrior:
Thanks @pierre!

PR #74, opened 2015-07-03
Add license file to project

jkeam:
Noticed that you had licensed individual files but not the project itself.


pierre:
Thanks for the report! Fixed by https://github.com/killbilling/recurly-java-library/commit/b6107c44c0267dc2c2b11d3e21587a0ad28fb9c3.

Issue #79, opened 2015-07-21
Transaction field mappings are off


jkeam:
There are a few transaction fields that were not mapped quite right.  For instance, the call to get a transaction: 

url: https://:subdomain.recurly.com/v2/transactions/:id

Returns the following result.  Take notice of cvv_result, avs_result.  These are being mapped to the PushTransaction, but as you can see here, these fields appear even on a normal non-webhook response.  Additionally, the avs_result_street and avs_result_postal fields are missing from the mapping.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<transaction href="https://your-subdomain.recurly.com/v2/transactions/a13acd8fe4294916b79aec87b7ea441f" type="credit_card">
  <account href="https://your-subdomain.recurly.com/v2/accounts/1"/>
  <invoice href="https://your-subdomain.recurly.com/v2/invoices/1108"/>
  <subscription href="https://your-subdomain.recurly.com/v2/subscriptions/17caaca1716f33572edc8146e0aaefde"/>
  <uuid>a13acd8fe4294916b79aec87b7ea441f</uuid>
  <action>purchase</action>
  <amount_in_cents type="integer">1000</amount_in_cents>
  <tax_in_cents type="integer">0</tax_in_cents>
  <currency>USD</currency>
  <status>success</status>
  <payment_method>credit_card</payment_method>
  <reference nil="nil"></reference>
  <source>subscription</source>
  <recurring type="boolean">true</recurring>
  <test type="boolean">true</test>
  <voidable type="boolean">true</voidable>
  <refundable type="boolean">true</refundable>
  <ip_address>127.0.0.1</ip_address>
  <cvv_result code="" nil="nil"></cvv_result>
  <avs_result code="" nil="nil"></avs_result>
  <avs_result_street nil="nil"></avs_result_street>
  <avs_result_postal nil="nil"></avs_result_postal>
  <created_at type="datetime">2011-06-27T12:34:56Z</created_at>
  <details>
    <account>
      <account_code>verena100</account_code>
      <first_name>Verena</first_name>
      <last_name>Example</last_name>
      <company nil="nil"></company>
      <email>verena@test.com</email>
      <billing_info type="credit_card">
        <first_name nil="nil"></first_name>
        <last_name nil="nil"></last_name>
        <address1 nil="nil"></address1>
        <address2 nil="nil"></address2>
        <city nil="nil"></city>
        <state nil="nil"></state>
        <zip nil="nil"></zip>
[...]

PR #80, opened 2015-07-21
Issue 79 - Fixed Transaction Mappings

jkeam:
This is for Issue https://github.com/killbilling/recurly-java-library/issues/79

A few of the Transaction fields are mapped incorrectly or are missing.

1.  cvv_result, avs_result were both being mapped to the PushTransaction, but needs to also appear in the regular Transaction object.  Both share the AbstractTransaction object, so I put the mapping in there.

2.  avs_result_street and avs_result_postal are missing from the mappings.  Since these are shared by both the PushTransaction and regular Transaction object, I put it inside the AbstractTransaction object.

3.  'source' field was being mapped to the Transaction object, but appears in a few webhooks.  B/c it is shared by PushTransaction also, I put this field inside AbstractTransaction.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/AbstractTransaction.java
+++ b/src/main/java/com/ning/billing/recurly/model/AbstractTransaction.java
@@ -18,9 +18,54 @@ package com.ning.billing.recurly.model;
 
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
+import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


jkeam:
Any thoughts on this?

bhelx:
@jkeam I have time to look at these tomorrow. Sorry for the delay. They both look like they should go in with little trouble.

jkeam:
yay thanks!!!

bhelx:
Thanks for the solid code with tests :)

PR #81, opened 2015-08-03
Updating docs links

csmb:
Recurly recently switched from `docs.recurly.com/api` to `dev.recurly.com` a. Navigating to an old docs link will redirect to the new docs site. These changes just make that new URL explicit.


pierre:
:beers: 

Thanks!

PR #82, opened 2015-08-17
Remove self-closing elements, eliminating jdk 1.8 javadoc complaints

easel:
Seemingly attempting to `mvn install` under jdk8 fails to build the documentation with lots of errors like this:

recurly-java-library/src/main/java/com/ning/billing/recurly/RecurlyJs.java:74: error: self-closing element not allowed
[ERROR] * <p/>

This patch fixes these errors.


pierre:
Thanks!

Issue #71, opened 2015-04-16
Setup travis for PR


pierre:
Integration tests could be run against a sandbox automatically.

bhelx:
This can be closed due to https://github.com/killbilling/recurly-java-library/pull/83

bhelx:
I'm going to look into this but I'm unsure the best way to get the integration tests running in Travis. Perhaps I can create a sandbox site and somehow set the keys only in Travis. I'd also need to look and see if api rate limiting will affect this.

pierre:
You can set private (*secure*) environment variables in Travis (I can set it up if you PM me the secret API key): you can then override the `script` entry in the `travis.yml` to pass them on the command line (`-D...`).

BTW, it would be great if we could have a sandbox with enterprise features enabled, so we could run the `enterprise` suite as well.

bhelx:
Yeah, good idea. I can setup whatever kind of sandbox and add all the feature flags to it. I noticed the integration tests don't work when you have line item refunds enabled but that's something I can fix later. Let me talk to some people about the rate limiting thing. I think it should be fine though.

pierre:
Any update on getting a sandbox just for Travis? Since we have lots of changes these days, this could be  quite useful.

bhelx:
@pierre this is on my task list for this week. I think it should be pretty quick. Expect a PR in the next couple days.

bhelx:
Closed with https://github.com/killbilling/recurly-java-library/pull/108

PR #83, opened 2015-09-14
Add Travis support

bhelx:
Starting to address https://github.com/killbilling/recurly-java-library/issues/71

This currently only runs unit tests. We'll need to discuss a way to run integration tests in Travis. Ideally we could privately set the credentials to a sandbox account.

@pierre Killbilling org needs to be enabled in Travis by an org admin. I believe this would be the url for you: https://travis-ci.org/profile/killbilling

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: java
+script:
+  - mvn clean test


pierre:
> Ideally we could privately set the credentials to a sandbox account.

:+1: Travis supports secret environment variables. Take a look at our [AvaTax .travis.yml](https://github.com/killbill/killbill-avatax-plugin/blob/master/.travis.yml#L8) for instance.

Would it be possible to have a sandbox account with entreprise features?

> Killbilling org needs to be enabled in Travis by an org admin

Just did it, will merge the PR to test it out.

pierre:
W00t! https://travis-ci.org/killbilling/recurly-java-library/builds/80303806

Issue #98, opened 2016-03-04
Incorrect RootElement for Accounts class


bhelx:
This should be `accounts` and not `plans` correct?

https://github.com/killbilling/recurly-java-library/blob/master/src/main/java/com/ning/billing/recurly/model/Accounts.java#L26

```java
@XmlRootElement(name = "plans")
public class Accounts extends RecurlyObjects<Account> {
   //...
}
```

I'm assuming it has not come up because it's only used for serialization and the clients never serialize a list of accounts? I'd be happy to send a fix for this.

\cc @pierre @sbrossie

pierre:
It looks like an [old typo](https://github.com/killbilling/recurly-java-library/commit/49569c71eab6e6306d1ec39828219c8a42db164f) indeed.

Just run the whole test suite with the fix to make sure we didn't miss anything :smile_cat:

bhelx:
Sure thing, already did. Will send a PR on monday morning.

bhelx:
Closed with https://github.com/killbilling/recurly-java-library/pull/99

PR #99, opened 2016-03-07
Fix root element for Accounts

bhelx:
Addresses issue discussed here:

https://github.com/killbilling/recurly-java-library/issues/98

Should have no functional change and pass all tests.


pierre:
:+1: 

Thanks!

PR #100, opened 2016-03-08
Fix test billing info expiration year

bhelx:
Fixes the billing info expiration year for the integration tests:

https://github.com/killbilling/recurly-java-library/blob/master/src/test/java/com/ning/billing/recurly/TestUtils.java#L99_L101

Changing to 2020 to give us some room.


pierre:
:+1:

PR #101, opened 2016-03-08
Fix failing integration test

bhelx:
I was seeing this integration test failing here:

https://github.com/killbilling/recurly-java-library/blob/master/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java#L159

The values of the objects were the same but their hashcodes were different. I narrowed this down to a missing hashcode override on the `VerificationResult` class. After adding that, the hashcodes for the two invoices were the same but it was still failing. It seems that `RecurlyObject` does not have an override for equals to check the hashcode so it's comparing the references.

I'm not sure if we want to override equals on `RecurlyObject` but in the mean time i've changed this to just assert that the hashcodes are the same as that is what I believe to be the original intent of this test. This at least fixes the test.

Do you have an opinion on this @pierre ? What was the original intent of overriding hashcode everywhere if you weren't also overriding equals?

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/AbstractTransaction.java
+++ b/src/main/java/com/ning/billing/recurly/model/AbstractTransaction.java
@@ -54,6 +54,13 @@ public class AbstractTransaction extends RecurlyObject {
             this.message = message;
         }
 
+        @Override
+        public int hashCode() {
+            int result = code != null ? code.hashCode() : 0;
+            result = 31 * result + (message != null ? message.hashCode() : 0);
+            return result;
+        }
+
         public static VerificationResult as(final Object object) {
             if (isNull(object)) {
                 return null;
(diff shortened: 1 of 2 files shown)


pierre:
The models should have both `hashCode` and `equals` [properly implemented](https://github.com/killbilling/recurly-java-library/blob/8f15de04589a30260677346b7068b23603373848/src/main/java/com/ning/billing/recurly/model/Account.java#L257-L343). If not, it's an oversight.

Happy to merge this but maybe we should do a pass at all models and implement the missing overrides?

bhelx:
How about I do a follow up PR for it today? I'd like to test some of these hash code methods as well.

pierre:
That sounds great!

bhelx:
Closing in favor of https://github.com/killbilling/recurly-java-library/pull/102

PR #102, opened 2016-03-09
Refactor and test code around equals and hashCode

bhelx:
As discussed here: https://github.com/killbilling/recurly-java-library/pull/101

This is still a work in progress

This code ensures that each model overrides `hashCode()`. It also makes sure not to use super's hashCode implementation because that was including references. We want to do a logical compare and not a reference compare. This code also removes the individual overrides of `equals` in favor of sharing one equals method on the `RecurlyObject` class.

It also adds a test case for each model and some helper methods to deterministically build random model data.

@pierre Let me know if you have any thoughts.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/AbstractAddOn.java
+++ b/src/main/java/com/ning/billing/recurly/model/AbstractAddOn.java
@@ -18,6 +18,7 @@
 package com.ning.billing.recurly.model;
 
 import javax.xml.bind.annotation.XmlElement;
+import com.google.common.base.Objects;
 
 public class AbstractAddOn extends RecurlyObject {
 
@@ -42,12 +43,8 @@ public class AbstractAddOn extends RecurlyObject {
 
     @Override
     public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
 
         final AbstractAddOn that = (AbstractAddOn) o;
 
@@ -60,6 +57,6 @@ public class AbstractAddOn extends RecurlyObject {
 
     @Override
     public int hashCode() {
-        return addOnCode != null ? addOnCode.hashCode() : 0;
+        return Objects.hashCode(addOnCode);
     }
 }
--- a/src/main/java/com/ning/billing/recurly/model/Account.java
+++ b/src/main/java/com/ning/billing/recurly/model/Account.java
@@ -17,6 +17,7 @@
 
 package com.ning.billing.recurly.model;
 
+import com.google.common.base.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -256,12 +257,8 @@ public class Account extends RecurlyObject {
 
     @Override
     public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
 
         final Account account = (Account) o;
 
@@ -283,7 +280,7 @@ public class Account extends RecurlyObject {
         if (companyName != null ? !companyName.equals(account.companyName) : account.companyName != null) {
             return false;
         }
-        if (createdAt != null ? !createdAt.equals(account.createdAt) : account.createdAt != null) {
+        if (createdAt != null ? createdAt.compareTo(account.createdAt) != 0 : account.createdAt != null) {
             return false;
         }
         if (email != null ? !email.equals(account.email) : account.email != null) {
@@ -322,23 +319,24 @@ public class Account extends RecurlyObject {
 
     @Override
     public int hashCode() {
-        int result = address != null ? address.hashCode() : 0;
-        result = 31 * result + (href != null ? href.hashCode() : 0);
-        result = 31 * result + (adjustments != null ? adjustments.hashCode() : 0);
-        result = 31 * result + (invoices != null ? invoices.hashCode() : 0);
-        result = 31 * result + (subscriptions != null ? subscriptions.hashCode() : 0);
-        result = 31 * result + (transactions != null ? transactions.hashCode() : 0);
-        result = 31 * result + (accountCode != null ? accountCode.hashCode() : 0);
-        result = 31 * result + (state != null ? state.hashCode() : 0);
-        result = 31 * result + (username != null ? username.hashCode() : 0);
-        result = 31 * result + (email != null ? email.hashCode() : 0);
-        result = 31 * result + (firstName != null ? firstName.hashCode() : 0);
-        result = 31 * result + (lastName != null ? lastName.hashCode() : 0);
-        result = 31 * result + (companyName != null ? companyName.hashCode() : 0);
-        result = 31 * result + (acceptLanguage != null ? acceptLanguage.hashCode() : 0);
-        result = 31 * result + (hostedLoginToken != null ? hostedLoginToken.hashCode() : 0);
-        result = 31 * result + (createdAt != null ? createdAt.hashCode() : 0);
-        result = 31 * result + (billingInfo != null ? billingInfo.hashCode() : 0);
-        return result;
+        return Objects.hashCode(
+                address,
+                href,
+                adjustments,
+                invoices,
+                subscriptions,
+                transactions,
+                accountCode,
+                state,
+                username,
+                email,
+                firstName,
+                lastName,
+                companyName,
+                acceptLanguage,
+                hostedLoginToken,
+                createdAt,
+                billingInfo
+        );
     }
 }
--- a/src/main/java/com/ning/billing/recurly/model/Invoice.java
+++ b/src/main/java/com/ning/billing/recurly/model/Invoice.java
@@ -22,6 +22,7 @@ import javax.xml.bind.annotation.XmlElementWrapper;
 import javax.xml.bind.annotation.XmlRootElement;
 
 import org.joda.time.DateTime;
+import com.google.common.base.Objects;
 
 @XmlRootElement(name = "invoice")
 public class Invoice extends RecurlyObject {
@@ -220,12 +221,8 @@ public class Invoice extends RecurlyObject {
 
     @Override
     public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
 
         final Invoice invoice = (Invoice) o;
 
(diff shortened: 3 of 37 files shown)


bhelx:
The build on travis is failing with:

```
[INFO] ------------------------------------------------------------------------

[INFO] BUILD FAILURE

[INFO] ------------------------------------------------------------------------

[INFO] Total time: 34.568 s

[INFO] Finished at: 2016-03-10T00:12:41+00:00

[INFO] Final Memory: 31M/139M

[INFO] ------------------------------------------------------------------------

[ERROR] Failed to execute goal org.apache.rat:apache-rat-plugin:0.7:check (default) on project recurly-java-library: Too many unapproved licenses: 1 -> [Help 1]

[ERROR] 

[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.

[ERROR] Re-run Maven using the -X switch to enable full debug logging.

[ERROR] 

[ERROR] For more information about the errors and possible solutions, please read the following articles:

[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException

The command "eval mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V" failed. Retrying, 2 of 3.
```

but runs fine locally

src/test/java/com/ning/billing/recurly/model/TestAddress.java
```diff
@@ -0,0 +1,24 @@
+package com.ning.billing.recurly.model;
```
pierre: You are missing the copyright header (which is why the Travis build is failing - you can check it via `mvn install -Ptravis`).

pierre:
> This is still a work in progress and PR 101 should be merged first.

I don't think we want #101. From http://docs.oracle.com/javase/6/docs/api/java/lang/Object.html#hashCode%28%29 :

> It is not required that if two objects are unequal according to the equals(java.lang.Object) method, then calling the hashCode method on each of the two objects must produce distinct integer results.

src/main/java/com/ning/billing/recurly/model/Account.java
```diff
@@ -255,72 +255,6 @@ public String toString() {
     }
 
     @Override
-    public boolean equals(final Object o) {
```
pierre: We cannot just rely on `hashCode` for `equals` (because of complex object types, like `DateTime`).

src/main/java/com/ning/billing/recurly/model/Account.java
```diff
@@ -255,72 +255,6 @@ public String toString() {
...
-        if (adjustments != null ? !adjustments.equals(account.adjustments) : account.adjustments != null) {
-            return false;
-        }
-        if (billingInfo != null ? !billingInfo.equals(account.billingInfo) : account.billingInfo != null) {
-            return false;
-        }
-        if (companyName != null ? !companyName.equals(account.companyName) : account.companyName != null) {
-            return false;
-        }
-        if (createdAt != null ? !createdAt.equals(account.createdAt) : account.createdAt != null) {
```
pierre: We should be using `compareTo` instead, to ignore the chronology.

src/main/java/com/ning/billing/recurly/model/Account.java
bhelx: What if we used the hashcode for everything but the datetimes? We could have models with datetimes call super then datetimes.

pierre: It sounds a bit error-prone IMHO.

Why not override both `equals` and `hashCode`? It's what I usually do (see [SubscriptionJson.java](https://github.com/killbill/killbill/blob/master/jaxrs/src/main/java/org/killbill/billing/jaxrs/json/SubscriptionJson.java#L460-L549) for example) and it can be (mostly) auto-generated by the IDE.

bhelx: The reason I was thinking of doing it this way is because it seems less error prone to only need one place to update when you add or remove fields (`hashcode` in this instance). I'm willing to defer that decision to you though. Do you know how would one go about automating this in IntelliJ?

pierre: Unfortunately, I don't know a way of automating this in IntelliJ. Alternatively, you could use Guava's [helpers](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Objects.html#hashCode(java.lang.Object...)).

I would just auto-generate them, because fields don't change that often. But I don't have a very strong opinion (just an old habit :smile_cat:).

bhelx:
Still need to do another pass at this next week with the changes we discussed.

bhelx:
@pierre I updated this PR in response to some of your comments

src/main/java/com/ning/billing/recurly/model/RecurlyObject.java
```diff
@@ -181,4 +181,12 @@ public static boolean isNull(@Nullable final Object object) {
     public void setRecurlyClient(final RecurlyClient recurlyClient) {
         this.recurlyClient = recurlyClient;
     }
+
+    @Override
+    public boolean equals(final Object o) {
```
bhelx: Hopefully this is still a sensible default. I feel like Object.equals is generally not what we want.

bhelx:
@pierre any way we can get this one in before the redemptions PR? Let me know if you have any more comments on it.

pierre:
Yup, yup, it's on my TODO list. This week was just quite busy! :sweat_smile:

I'll set some time aside today to review it again.

src/main/java/com/ning/billing/recurly/model/AbstractAddOn.java
```diff
@@ -18,6 +18,7 @@
 package com.ning.billing.recurly.model;
 
 import javax.xml.bind.annotation.XmlElement;
+import java.util.Objects;
```
pierre: `java.util.Objects` is 1.7+ only unfortunately. Do you have lots of enterprise users? Most of them usually still run 1.6 :crying_cat_face:

src/test/java/com/ning/billing/recurly/model/TestAddOns.java
```diff
@@ -17,10 +17,16 @@
 
 package com.ning.billing.recurly.model;
 
+import com.ning.billing.recurly.TestUtils;
 import org.joda.time.DateTime;
+import org.omg.PortableInterceptor.SYSTEM_EXCEPTION;
```
pierre: Extra import.

src/test/java/com/ning/billing/recurly/TestUtils.java
```diff
@@ -17,61 +17,166 @@
 
 package com.ning.billing.recurly;
 
+import java.util.Date;
+import java.util.Random;
 import java.util.UUID;
 
+import com.ning.billing.recurly.model.*;
```
pierre: Wildcard import.

src/main/java/com/ning/billing/recurly/model/Subscription.java
```diff
@@ -389,26 +383,27 @@ public boolean equals(final Object o) {
 
     @Override
     public int hashCode() {
-        int result = account != null ? account.hashCode() : 0;
-        result = 31 * result + (plan != null ? plan.hashCode() : 0);
-        result = 31 * result + (uuid != null ? uuid.hashCode() : 0);
-        //result = 31 * result + (state != null ? state.hashCode() : 0);
```
pierre: Odd that state was left out? We should re-introduce it.

src/main/java/com/ning/billing/recurly/model/RecurlyUnitCurrency.java
```diff
@@ -326,21 +323,21 @@ public boolean equals(final Object o) {
...
-        result = 31 * result + (unitAmountNZD != null ? unitAmountNZD.hashCode() : 0);
-        result = 31 * result + (unitAmountPLN != null ? unitAmountPLN.hashCode() : 0);
-        result = 31 * result + (unitAmountSGD != null ? unitAmountSGD.hashCode() : 0);
-        result = 31 * result + (unitAmountSEK != null ? unitAmountSEK.hashCode() : 0);
-        result = 31 * result + (unitAmountCHF != null ? unitAmountCHF.hashCode() : 0);
-        result = 31 * result + (unitAmountZAR != null ? unitAmountZAR.hashCode() : 0);
-        return result;
+        return Objects.hash(
+                unitAmountUSD,
+                unitAmountAUD,
```
pierre: Missing CAD.

src/main/java/com/ning/billing/recurly/model/Invoice.java
```diff
@@ -280,21 +278,21 @@ public boolean equals(final Object o) {
...
-        result = 31 * result + (transactions != null ? transactions.hashCode() : 0);
-        return result;
+        return Objects.hash(
+                account,
+                uuid,
+                state,
+                invoiceNumber,
+                poNumber,
+                vatNumber,
+                subtotalInCents,
```
pierre: Missing totalInCents.

pierre:
Looking good overall. Besides a few typo, and maybe changing Java 1.7 Objects to Guava Objects, :shipit:!

src/main/java/com/ning/billing/recurly/model/AbstractAddOn.java
bhelx: Well, I don't have the numbers unfortunately. I have a strong feeling that some do, but they will need to upgrade by this summer anyway due to TLS restrictions. We are going to need to add a 1.7 minimum anyway, we could do it now? 

Actually according to this SO article: http://stackoverflow.com/questions/10500511/how-to-find-what-ssl-tls-version-is-used-in-java

The IBM JRE6 supports TLS 1.1 and 1.2. So I'm not sure what way forward we should go.

bhelx: Oh, I can change to Guava Objects?

pierre: > We are going to need to add a 1.7 minimum anyway

I'm not sure we should if it's just because of the TLS upgrade (http://stackoverflow.com/questions/33364100/how-to-use-tls-1-2-in-java-6).

Also, are you sure you will force an upgrade on all endpoints (it makes sense for the ones in PCI scope, but not for others, like the ones dealing with subscription management). It can be a very hard upgrade on legacy deployments...

> Oh, I can change to Guava Objects?
 
Yup! http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Objects.html (it's probably going to be deprecated in favor of the built-in one, but we can still use it for now).

src/main/java/com/ning/billing/recurly/model/Invoice.java
bhelx: Good catch, added that below

src/main/java/com/ning/billing/recurly/model/Subscription.java
bhelx: I'm not sure, I was going to leave it the same but i re-introduced it below

src/test/java/com/ning/billing/recurly/model/TestAddress.java
bhelx: This should be taken care of now

bhelx:
Okay, updated from comments and changed to Guava Objects.

src/test/java/com/ning/billing/recurly/model/TestAddOns.java
pierre: Bump!

pierre:
ack - looks like there are still one extra import (`SYSTEM_EXCEPTION`).

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -67,8 +67,9 @@
 
     @BeforeMethod(groups = {"integration", "enterprise"})
     public void setUp() throws Exception {
-        final String apiKey = System.getProperty(KILLBILL_PAYMENT_RECURLY_API_KEY);
-        String subDomainTemp = System.getProperty(KILLBILL_PAYMENT_RECURLY_SUBDOMAIN);
+        final String apiKey = "a0f54c9dc5a84b97be1442e630cda2cb";
```
pierre: :warning:

bhelx: Regenerated. Thankfully only a sandbox site :)

pierre: It happens ;-)

pierre:
:tada:

bhelx:
Thanks for your time @pierre :)

Issue #97, opened 2016-03-03
Coupon Redemptions 


cbsi-sl:
Hi, 

Right now the method getCouponRedemptionByAccount() seems to hit this Recurly endpoint:
https://:subdomain.recurly.com/v2/accounts/:account_code/redemption

Recurly seems to also support listing out multiple redemptions, including redemptions that have naturally expired:
https://:subdomain.recurly.com/v2/accounts/:account_code/redemptions

May I ask if the multiple redemptions response will be supported?

thanks.

bhelx:
@cbsi-sl I'll put it on my calendar for next week. I need to investigate the level of effort to required to add it to this library. I think it should be pretty straight forward.

cbsi-sl:
Thanks for helping with this Benjamin!

bhelx:
Closed with https://github.com/killbilling/recurly-java-library/pull/104

PR #104, opened 2016-03-15
Support multiple coupon redemptions

bhelx:
This PR is in reference to issue https://github.com/killbilling/recurly-java-library/issues/97

It's still a work in progress but I wanted to open it up to comments.

A few comments:

* Since we are now using Recurly API v2.1 we may want to use a major version bump when deploying.
* For the integration tests to run you'll need to set support for multiple coupon redemptions on your `Coupon > Settings` page 

\cc @pierre @cbsi-sl

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -33,10 +33,6 @@ import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import javax.xml.bind.DatatypeConverter;
 
-import org.joda.time.DateTime;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.ning.billing.recurly.model.Account;
 import com.ning.billing.recurly.model.Accounts;
 import com.ning.billing.recurly.model.AddOn;
@@ -55,6 +51,7 @@ import com.ning.billing.recurly.model.RecurlyAPIError;
 import com.ning.billing.recurly.model.RecurlyObject;
 import com.ning.billing.recurly.model.RecurlyObjects;
 import com.ning.billing.recurly.model.Redemption;
+import com.ning.billing.recurly.model.Redemptions;
 import com.ning.billing.recurly.model.RefundOption;
 import com.ning.billing.recurly.model.Subscription;
 import com.ning.billing.recurly.model.SubscriptionUpdate;
@@ -62,6 +59,11 @@ import com.ning.billing.recurly.model.SubscriptionNotes;
 import com.ning.billing.recurly.model.Subscriptions;
 import com.ning.billing.recurly.model.Transaction;
 import com.ning.billing.recurly.model.Transactions;
+
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.ning.http.client.AsyncHttpClient;
 import com.ning.http.client.AsyncHttpClientConfig;
 import com.ning.http.client.Response;
@@ -80,6 +82,7 @@ public class RecurlyClient {
 
     public static final String RECURLY_DEBUG_KEY = "recurly.debug";
     public static final String RECURLY_PAGE_SIZE_KEY = "recurly.page.size";
+    public static final String RECURLY_API_VERSION = "2.1";
 
     private static final Integer DEFAULT_PAGE_SIZE = 20;
     private static final String PER_PAGE = "per_page=";
@@ -760,7 +763,7 @@ public class RecurlyClient {
     }
 
     /**
-     * Lookup a coupon redemption on an invoice.
+     * Lookup the first coupon redemption on an account.
      *
      * @param accountCode recurly account id
      * @return the coupon redemption for this account on success, null otherwise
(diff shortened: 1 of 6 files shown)


src/main/java/com/ning/billing/recurly/model/AbstractTransaction.java
```diff
@@ -54,6 +54,13 @@ public void setMessage(final String message) {
             this.message = message;
         }
 
+        @Override
+        public int hashCode() {
```
bhelx: This was snuck into to fix the integration tests. Will remove.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -790,6 +790,16 @@ public void deleteCouponRedemption(final String accountCode) {
         doDELETE(Accounts.ACCOUNTS_RESOURCE + "/" + accountCode + Redemption.REDEMPTION_RESOURCE);
     }
 
+    /**
+     * Deletes a specific redemption.
+     *
+     * @param accountCode recurly account id
+     * @param Redemption recurly coupon redemption
+     */
+    public void deleteCouponRedemption(final String accountCode, final Redemption redemption) {
```
bhelx: This may be easier to use if we pass in the uuid instead of a `Redemption` object

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -33,35 +33,11 @@
 import javax.annotation.Nullable;
 import javax.xml.bind.DatatypeConverter;
 
+import com.ning.billing.recurly.model.*;
```
pierre: Personally, I dislike wildcard imports (harder to figure out what's being imported).

src/main/java/com/ning/billing/recurly/RecurlyClient.java
pierre: :+1: on using a UUID instead.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -930,6 +940,7 @@ private void doDELETE(final String resource) {
         final Response response = builder.addHeader("Authorization", "Basic " + key)
                                          .addHeader("Accept", "application/xml")
                                          .addHeader("Content-Type", "application/xml; charset=utf-8")
+                                         .addHeader("X-Api-Version", RECURLY_API_VERSION)
```
pierre: What's the impact of setting this?

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -22,6 +22,7 @@
 import java.util.List;
 import java.util.Set;
 
+import com.ning.billing.recurly.model.*;
```
pierre: Same remark.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: I believe IntelliJ did this automatically. I'll set it back and see if I can change the IntelliJ settings.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: It's not a fully documented feature yet, but this is required for using the multiple redemptions endpoints:

https://dev.recurly.com/docs/getting-started#additional-headers

This shouldn't effect anything beyond this PR but I should run through the changes on our end and double check.

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -155,8 +135,8 @@ public void testRemoveSubscriptionAddons() throws Exception {
             // Refetch the invoice using the getInvoice method
             final int invoiceID = subInvoice.getInvoiceNumber();
             final Invoice gotInvoice = recurlyClient.getInvoice(invoiceID);
-            
-            Assert.assertEquals(subInvoice,gotInvoice);
+
+            Assert.assertEquals(subInvoice.hashCode(), gotInvoice.hashCode());
```
bhelx: Again, ignore this line. Will be removed.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
pierre: Oh, I see. Just specify 999 in Editor > Code Style > Java:

image

See also https://github.com/killbill/killbill/blob/master/.idea/codeStyleSettings.xml#L9-L10.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
pierre: Ack. Maybe `RECURLY_API_VERSION` should be a System property?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: Ah, cool. Thanks for finding it! I've been using vim for the past 4 years or so. Trying to get familiar with the mysteries of these IDEs.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: Well, I don't think it's something we'd want people to be able to change. It should travel with the code because the API has a different interface depending on the Recurly version you use, thus this version number is inseparable from the code. So if someone changed the value, the code would possibly be broken in some spots. Right now we are mostly using it as a way to move people forward on new features without breaking compatibility for older merchants. In general it's something that should be hidden from the merchants and they should only know about what version of the client library they are using.

bhelx:
@pierre Updated as per your comments

bhelx:
This has been rebased and updated.

Issue #96, opened 2016-03-02
TLS 1.1 or later


jtvgithub:
I am getting following warning in Recurly console. Please help -

**Important Notice: We have detected API connections to this Recurly subdomain using TLS 1.0. For security reasons these connections must be upgraded to TLS 1.1 or later by June 30th, 2016.**

bhelx:
Hey @jtvgithub

Can you tell us what Operating System you are using and which version of Java you are using?

jtvgithub:
After updating Java, this problem is solved.

bhelx:
@jtvgithub Glad it's working for you. Going to close this ticket.

Could you let us know which version you were on and what you updated to? If someone comes searching it would be helpful if they could find some information.

bhelx:
@pierre We are transitioning everyone over to TLS1.1 and TLS1.2 over the summer for PCI compliance reasons (TLS1.0 being deprecated). As far as I can read, Java7 supports these protocols only if you explicitly enable them in the VM and Java8 enables them by default. Do you know if we need to change anything with `AsyncHttpClient` so that it can use the most up to date protocol?

bhelx:
@jtvgithub Did you update from 7 to 8?

pierre:
@bhelx I would guess it's a JVM configuration issue, AHC will simply delegate protocol management to the JVM.

That being said, if you want to enforce the TLS version, this might be possible via [configuration](https://github.com/killbill/killbill-plugin-framework-java/blob/7fac2b75992de3f244e530ce1e8db15a3823cbbd/src/main/java/org/killbill/billing/plugin/util/http/HttpClient.java#L87).

bhelx:
Okay great, I just want to get a plan before we close this issue again. I'm gonna do a couple experiments and some reading and get back.

Will also be updating our TLS 1.0 deprecation document with the plan:

https://dev.recurly.com/v2.0/page/tls-v10-deprecation

bhelx:
@pierre From what I can tell, both 1.7 and 1.8 support TLS 1.1 and 1.2. Version 1.8 will do the negotiation properly on it's own but on version 1.7 we must pass in the protocol to the `getInstance` method on the SSLContext. Would you recommend that I make a PR against your killbill-plugin-framework-java project? What should the configuration look like? Should we just set the protocol to 1.2 automatically if the project is using 1.7 or should there be some other way for the client to configure the protocol?

bhelx:
[...]

PR #110, opened 2016-04-07
Enforce TLS 1.2

bhelx:
Related to discussion in issue https://github.com/killbilling/recurly-java-library/issues/96

Set's the protocol to TLS version 1.2 or whatever is passed.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -24,6 +24,8 @@ import java.math.BigDecimal;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.security.GeneralSecurityException;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
 import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Scanner;
@@ -1069,8 +1071,11 @@ public class RecurlyClient {
             // Don't limit the number of connections per host
             // See https://github.com/ning/async-http-client/issues/issue/28
             builder.setMaxConnectionsPerHost(-1);
-            builder.setSSLContext(SslUtils.getInstance().getSSLContext(false));
-        } catch (GeneralSecurityException e) {
+            builder.setSSLContext(SslUtils.getInstance().getSSLContext());
+        } catch (NoSuchAlgorithmException e) {
+            // throw a recurly exception here?
+            e.printStackTrace();
+        } catch (KeyManagementException e) {
             // throw a recurly exception here?
             e.printStackTrace();
         } finally {
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -23,6 +23,7 @@ import java.io.Reader;
 import java.math.BigDecimal;
 import java.net.URL;
 import java.nio.charset.Charset;
+import java.security.GeneralSecurityException;
 import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Scanner;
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/util/http/SslUtils.java
```diff
@@ -0,0 +1,70 @@
...
+
+        public void checkServerTrusted(final java.security.cert.X509Certificate[] certs, final String authType) {
+        }
+    }
+
+    private SSLContext looseTrustManagerSSLContext = looseTrustManagerSSLContext();
+
+    private SSLContext looseTrustManagerSSLContext() {
+        try {
+            final SSLContext sslContext = SSLContext.getInstance("TLSv1.2");
```
bhelx: This doesn't appear to be working. I'm still seeing TLS1.0 on the server side when running this in Travis.

Also, wondering how we can make this configurable.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -1061,11 +1063,19 @@ private void closeStream(final InputStream in) {
...
+
+        try {
+            // Don't limit the number of connections per host
+            // See https://github.com/ning/async-http-client/issues/issue/28
+            builder.setMaxConnectionsPerHost(-1);
+            builder.setSSLContext(SslUtils.getInstance().getSSLContext(false));
+        } catch (GeneralSecurityException e) {
+            // throw a recurly exception here?
+            e.printStackTrace();
+        } finally {
```
bhelx: I imagine we want to turn this into a standard RecurlyError of some sort

src/main/java/com/ning/billing/recurly/util/http/SslUtils.java
pierre: > This doesn't appear to be working

It looks like your are calling `SslUtils.getInstance().getSSLContext(false)` which makes it return `SSLContext.getDefault();` (see below).

BTW, I don't think we want `LooseTrustManager` anyways, which is to support self-signed (and invalid) certificates.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
pierre: I'm not sure since it's a client setup issue, not a server-side issue. I would re-throw `GeneralSecurityException` (it changes the `open` API but at least it's not ambiguous).

bhelx:
Okay, this appears to be working at least

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: Okay, I was concerned that would change all the method signatures but it looks like it only changes the setup like yous stated.

src/main/java/com/ning/billing/recurly/util/http/SslUtils.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public static SslUtils getInstance() {
+        return SingletonHolder.instance;
+    }
+
+    public SSLContext getSSLContext() throws NoSuchAlgorithmException, KeyManagementException {
+        if (context != null) return this.context;
+
+        context = SSLContext.getInstance("TLSv1.2");
+        context.init(null, null, null);
```
bhelx: What are the implications of these being null? I assume it has safe defaults?

bhelx:
@pierre It's now configurable via a system property and defaults to v1.2. Let me know if this is what you were thinking.

pierre:
Yup, looking good!

I would also add a note in the README, to help people debug when upgrading, it case it breaks on older installations.

bhelx:
Thanks @pierre !

I'll do a follow up PR for the README change

pierre:
Thank **you** for teaching me about [Cargo cult programming](https://en.wikipedia.org/wiki/Cargo_cult_programming)! :joy_cat:

bhelx:
It's been known to happen on Fridays :)

PR #111, opened 2016-04-12
Add TLS property documentation

bhelx:
This updates the README to explain how to use the `tlsProtocol` property.

I also needed to fix the spelling on those *_PROTOCOL_* constants.

@pierre What do you think of adding another section in the readme that states we only support Java 7 and above?

### DIFF
--- a/src/main/java/com/ning/billing/recurly/util/http/SslUtils.java
+++ b/src/main/java/com/ning/billing/recurly/util/http/SslUtils.java
@@ -24,8 +24,8 @@ import javax.net.ssl.SSLContext;
 
 public class SslUtils {
 
-    private static final String TLS_PROTCOL_KEY = "killbill.payment.recurly.tlsProtocol";
-    private static final String TLS_PROTCOL_DEFAULT = "TLSv1.2";
+    private static final String TLS_PROTOCOL_KEY = "killbill.payment.recurly.tlsProtocol";
+    private static final String TLS_PROTOCOL_DEFAULT = "TLSv1.2";
     private SSLContext context;
 
     private static class SingletonHolder {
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/util/http/SslUtils.java
```diff
@@ -24,8 +24,8 @@
 
 public class SslUtils {
 
-    private static final String TLS_PROTCOL_KEY = "killbill.payment.recurly.tlsProtocol";
-    private static final String TLS_PROTCOL_DEFAULT = "TLSv1.2";
+    private static final String TLS_PROTOCOL_KEY = "killbill.payment.recurly.tlsProtocol";
```
pierre: :+1: Oops, missed this too.

pierre:
> What do you think of adding another section in the readme that states we only support Java 7 and above?

Is that actually the case? The code is compatible with Java 7 and TLS 1.1 is supported since [JDK 6u111](http://www.oracle.com/technetwork/java/javase/overview-156328.html).

I'm happy to drop support for Java 6 if no-one cares about it anymore but could we check it first? Could you maybe extract some stats using the [User-Agent](https://github.com/killbilling/recurly-java-library/pull/59/files)?

bhelx:
Hmm, I was under the impression that > TLS1.0 support was only available on the IBM supported version of jdk6.

I'll run a query in our system to see what Java versions people are using and get back.

bhelx:
I ran a query against all KillBill useragents in the system. All of them are 1.7 and above.

pierre:
Perfect! 1.7+ it is then, new goodies here we go... :smiley_cat:

PR #113, opened 2016-04-19
Support for trial extension coupons

bhelx:
This feature is in production but is not available for every site yet. I've added the feature to our sandbox site. It's supposed to be available on Thursday.

The feature is a special type of coupon that allows an extension of your subscription trial period.

The API changes are only a few added fields.

\cc @pierre

### DIFF
--- a/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
+++ b/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
@@ -47,6 +47,7 @@ import com.ning.billing.recurly.model.Transactions;
 
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.joda.time.LocalDateTime;
 import org.joda.time.Minutes;
 import org.testng.Assert;
 import org.testng.annotations.AfterMethod;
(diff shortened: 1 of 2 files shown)


src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -1112,4 +1112,45 @@ public void testRedeemCoupon() throws Exception {
...
+        couponData.setName("apitrialext");
+        couponData.setCouponCode("apitrialext");
+        couponData.setTrialExtensionAmount(3);
+        couponData.setTrialExtensionUnit("month");
+        couponData.setDiscountType("trial_extension");
+
+        try {
+            final Account account = recurlyClient.createAccount(accountData);
+            final Plan plan = recurlyClient.createPlan(planData);
+            final Coupon coupon = recurlyClient.createCoupon(couponData);
```
pierre: Is the coupon automatically redeemed? Or should it be applied to the subscription explicitly?

bhelx: The `coupon_code` needs to be added to the `subscriptionData` below. The test should then check that the coupon was applied. I'll update this when the feature is ready :)

bhelx:
@pierre Can you give this another look when you get the chance? The feature is in production now.

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -1112,4 +1113,50 @@ public void testRedeemCoupon() throws Exception {
...
+            subscriptionData.setCouponCode(coupon.getCouponCode());
+
+            subscriptionData.setPlanCode(plan.getPlanCode());
+            subscriptionData.setAccount(accountData);
+            subscriptionData.setCurrency(CURRENCY);
+            subscriptionData.setUnitAmountInCents(1242);
+
+            final Subscription subscription = recurlyClient.createSubscription(subscriptionData);
+            Assert.assertNotNull(subscription);
+            Assert.assertEquals(subscription.getTrialEndsAt().getMonthOfYear(), now.getMonthOfYear() + 3);
```
bhelx: This is meant to check that the trial was extended by 3 months

pierre:
👍

PR #115, opened 2016-05-02
Change name of free trial api attributes

bhelx:
@pierre Sorry to expose you to this.

The team responsible for this feature decided to change the name of the feature and thus these API fields. They changed it from trial extension to free trial. This should not affect anyone as the feature is not being used by anyone yet. I have been assured this is the final API. Sorry again.


pierre:
No problem, naming is hard ;-)

bhelx:
thanks @pierre !

Issue #116, opened 2016-05-06
Provide HTTP status code in RecurlyApiExceptions


steelThread:
Would be beneficial for clients of this lib to include the reculy response http status code in the RecurlyApiExceptions thrown from RecurlyClient.  I'm trying to setup some retry logic for some situations, 500 level statuses namely, and I can't find a deterministic way for determining when I encounter this situation.

bhelx:
Hey @steelThread

Which situations do you with to retry on? I may be able to help you determine them without this feature. If not we can consider adding something in to help.

steelThread:
@bhelx - Thanks for getting back so quickly.  I have a few things here, I want to deal with intermittent network errors on our side.   Looks like I get back a null when the request fails with a ConnectException (ExecutionException really).  Think I have that covered.  The other conditions would be any of the 500 statuses, with these is there way to determine from the RecurlyApiException what the error was?  Do they return an xml doc that gets mapped to the RecurlyAPIError?  Thanks in advance for the help.

steelThread:
@bhelx - any ideas here?

jaaknt:
+1 to idea, to include http response status code in RecurlyAPIException if
response.getStatusCode() >= 300

bhelx:
@steelThread Sorry this fell of my radar. Our API errors have descriptive information in them but there are situations where you may not get it (we've been having recent Cloudflare HTTP level errors for example). 

For that reason, I'm fine with adding a the state code on the ApiError. Sounds like it would be a good addition. Unless one of y'all wants to take the task I am happy to add it.

steelThread:
That would be great @bhelx.

bhelx:
@steelThread This is on my queue. Will link back here when the PR is published.

qualidafial:
Can we do something about the null return value on a 500? Seems like this should also result in a `RecurlyAPIException`.

pierre:
@qualidafial Did you try https://github.com/killbilling/recurly-java-library/pull/120/files?

bhelx:
I believe this was closed with #120 and #123

Please reopen if this is not the case.

PR #120, opened 2016-05-31
Add HTTP status to exception

bhelx:
Meant to address https://github.com/killbilling/recurly-java-library/issues/116

\cc @steelThread

You should be able to get the response code:

```java
apiException.getRecurlyError().getHttpStatusCode();
```

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -1010,12 +1010,13 @@ public class RecurlyClient {
                     }
                     throw new TransactionErrorException(errors);
                 } else {
-                    RecurlyAPIError recurlyError = null;
+                    RecurlyAPIError recurlyError = new RecurlyAPIError();
                     try {
                         recurlyError = xmlMapper.readValue(payload, RecurlyAPIError.class);
                     } catch (Exception e) {
                         log.debug("Unable to extract error", e);
                     }
+                    recurlyError.setHttpStatusCode(response.getStatusCode());
                     throw new RecurlyAPIException(recurlyError);
                 }
             }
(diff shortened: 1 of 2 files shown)


jaaknt:
I fixed in my fork already and problem is that recurlyError can be null
                    RecurlyAPIError recurlyError = null;
                    try {
                        recurlyError = xmlMapper.readValue(payload, RecurlyAPIError.class);
                    } catch (Exception e) {
                        log.debug("Unable to extract error", e);
                        if (recurlyError == null) {
                           recurlyError = new RecurlyAPIError();
                        }
                    }
                    recurlyError.setStatusCode(response.getStatusCode());
                    throw new RecurlyAPIException(recurlyError);

bhelx:
@jaaknt does that change work for you?

@pierre any opinions?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -1010,12 +1010,13 @@ private void doDELETE(final String resource) {
                     }
                     throw new TransactionErrorException(errors);
                 } else {
-                    RecurlyAPIError recurlyError = null;
+                    RecurlyAPIError recurlyError = new RecurlyAPIError();
```
bhelx: This has the consequence that the recurlyError is never null but I think it should be okay.

src/main/java/com/ning/billing/recurly/model/RecurlyAPIError.java
```diff
@@ -57,12 +59,17 @@ public void setDetails(final String details) {
         this.details = details;
     }
 
+    public void setHttpStatusCode(final int httpStatusCode) { this.httpStatusCode = httpStatusCode; }
```
pierre: Nit: current code style has these expanded.

pierre:
👍

bhelx:
Thanks @pierre 

fixed formatting and rebased

steelThread:
Any idea when this change will be pushed out to maven central?

bhelx:
Thanks @pierre 👍

PR #127, opened 2016-07-18
Update to API version 2.2

bhelx:
This updates the HTTP API to version 2.2.

This brings the `updated_at` field and some other features which I can add in a piecemeal fashion later. This should be the minimum required to get some new features into this library.


pierre:
👍

PR #128, opened 2016-07-21
Add support to retrieve invoice pdf using invoice number

jaypatel11:
-- Added support to retrieve invoice pdf as byte array
-- Added logic for sending and receiving response when content type is application/pdf

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -17,13 +17,10 @@
 
 package com.ning.billing.recurly;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.Reader;
+import java.io.*;
 import java.math.BigDecimal;
 import java.net.URL;
 import java.nio.charset.Charset;
-import java.security.GeneralSecurityException;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.NoSuchElementException;
@@ -557,6 +554,18 @@ public class RecurlyClient {
     }
 
     /**
+     * Fetch invoice pdf
+     * <p>
+     * Returns the invoice pdf as a byte array
+     *
+     * @param invoiceId Recurly Invoice ID
+     * @return the invoice pdf as a byte array
+     */
+    public byte[] getInvoicePdf(final Integer invoiceId) {
+        return doGETPdf(Invoices.INVOICES_RESOURCE + "/" + invoiceId);
+    }
+
+    /**
      * Lookup an account's invoices
      * <p>
      * Returns the account's invoices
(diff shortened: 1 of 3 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -900,14 +918,53 @@ public Invoice fetchInvoice(final String recurlyToken) {
...
+        } catch (ExecutionException e) {
+            log.error("Execution error", e);
+            return null;
+        } catch (IOException e) {
+            log.error("Error retrieving response body", e);
+            return null;
+        }
+
+        if (response.getStatusCode() != 200) {
+            RecurlyAPIError recurlyAPIError = new RecurlyAPIError();
```
jaypatel11: When the content type is application/pdf, recurly does not return any errors with it. I tested it by trying to retrieve a pdf for an invalid invoice. The response content is text/html and the content is an empty string

bhelx:
@jaypatel11 thanks a lot for adding this feature!

I'm curious as to the use of pdfbox, our clients normally just expose the pdf as raw bytes and let the application developer decide what to do with it. Is there a reason we need another dependency here?

jaypatel11:
@bhelx The pdfbox is used for unit testing the retrieved pdf in TestRecurlyClient

src/main/java/com/ning/billing/recurly/RecurlyClient.java
pierre: This is something maybe @bhelx could help fix on the server side. Would it be possible to return 404 in that case?

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -716,6 +718,52 @@ public void testCreateAndCloseInvoices() throws Exception {
     }
 
     @Test(groups = "enterprise")
+    public void testCreateInvoiceAndRetrieveInvoicePdf() throws Exception {
```
pierre: Is it only available for enterprise users?

jaypatel11: Good catch. This is available for non enterprise as well.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
jaypatel11: When retrieving an invoice pdf for a non-existing invoice id, the content-type is "text/html" and since the client expects "application/pdf" we get a 406 error(Non acceptable response format).

jaypatel11: Yeah, 404 sounds reasonable for that case

jaypatel11: I just tested the getInvoice() with invalid invoice Id. I got this:
RecurlyAPIError{description='Couldn't find Invoice with invoice_number = 1818181', symbol='not_found', details='null', httpStatusCode='404'}

bhelx: I would think HTTP level error codes should all be the same as any other content type.

bhelx: I will check.

jaypatel11:
@bhelx Any update on this?

bhelx:
I haven't had a chance to look at the 404 issue on the server side, but the code looks good here. @pierre any more suggestions?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -17,13 +17,10 @@
 
 package com.ning.billing.recurly;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.Reader;
+import java.io.*;
```
pierre: Wildcard imports.

jaypatel11: addressed

PR #129, opened 2016-07-25
Update to API version 2.3

bhelx:
This updates us to API version 2.3.

The biggest change here is the deprecation of going backwards in pagination. We have also added support for getting the account balance which was the last feature advantage that v1 had over v2.

* Add revenue recognition fields
* Remove support for `prev` in pagination
* Add `AccountBalance`

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/AddOn.java
+++ b/src/main/java/com/ning/billing/recurly/model/AddOn.java
@@ -42,6 +42,9 @@ public class AddOn extends AbstractAddOn {
     @XmlElement(name = "unit_amount_in_cents")
     private RecurlyUnitCurrency unitAmountInCents;
 
+    @XmlElement(name = "revenue_schedule_type")
+    private RevenueScheduleType revenueScheduleType;
+
     @XmlElement(name = "created_at")
     private DateTime createdAt;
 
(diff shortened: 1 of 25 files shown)


src/main/java/com/ning/billing/recurly/model/AddOn.java
```diff
@@ -72,6 +75,14 @@ public void setDefaultQuantity(final Object defaultQuantity) {
         this.defaultQuantity = integerOrNull(defaultQuantity);
     }
 
+    public RevenueScheduleType getRevenueScheduleType() {
+        return revenueScheduleType;
+    }
+
+    public void setRevenueScheduleType(final String revenueScheduleType) {
+        this.revenueScheduleType = RevenueScheduleType.valueOf(revenueScheduleType.toUpperCase());
```
bhelx: I feel like there may be a better way to do this with enums that I don't know about.

src/main/java/com/ning/billing/recurly/model/RevenueScheduleType.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+
+package com.ning.billing.recurly.model;
+
+public enum RevenueScheduleType {
```
bhelx: Perhaps this enum shouldn't stand on it's own but instead be in some kind of `RevenueRecognition` namespace class. Not sure if we have a place to put an enum that is shared by many models right now.

pierre: I think it's fine the way you have it. If more reports are exposed moving forward, maybe we could then re-think it?

BTW, I couldn't find any documentation on these fields - is this a public API? If not, could you briefly describe what they mean?

pierre:
👍

src/main/java/com/ning/billing/recurly/model/RevenueScheduleType.java
bhelx: Sure, I'll update the description.

We are currently versioning out our docs and version 2.3 will be listed soon.

bhelx: Here are the revenue recognition feature docs: https://docs.recurly.com/docs/revenue-recognition

PR #91, opened 2015-11-02
using Jackson 2.5 and up causes issues when interacting with reculry

mzouitni:
Upgrading to Jackson 2.5 or higher causes recurly api calls to stop working. An example of a failed create subscription call:
```
<?xml version="1.0" encoding="UTF-8"?>
<error>
  <symbol>invalid_xml</symbol>
  <description>The provided XML was invalid.</description>
  <details>Unacceptable tag &lt;state&gt;</details>
</error>
```
This is the actual xml generated by recurly java library:
```
<subscription><unit_amount_in_cents/><quantity/><plan_code>full</plan_code><account><address/><account_code>xxxxxxxxxx</account_code><state/><username/><email/><first_name/><last_name/><company_name/><accept_language/><hosted_login_token/><created_at/><billing_info><account/><first_name/><last_name/><company/><address1/><address2/><city/><state/><zip/><country/><phone/><vat_number/><ip_address/><ip_address_country/><card_type/><year/><month/><first_six/><last_four/><number/><verification_value/><token_id>xyyxyxxyxyxy</token_id></billing_info></account><invoice/><plan/><uuid/><state/><currency>USD</currency><activated_at/><canceled_at/><expires_at/><current_period_started_at/><current_period_ends_at/><trial_started_at/><trial_ends_at/><pending_subscription/><starts_at/><collection_method/><net_terms/><coupon_code/><po_number/><terms_and_conditions/><customer_notes/><first_renewal_date/><bulk/></subscription>
```

Prior to the upgrade the actual xml doesn't include empty or null values like the example below:
```
<subscription><plan_code>plan</plan_code><account><account_code>xxxxxxxxxx</account_code><billing_info><token_id>xyyxyxxyxyxy</token_id></billing_info></account><currency>USD</currency></subscription>
```
Adding the below line to the RecurlyObject.java in newXmlMapper() method, fixes the problem.
```
xmlMapper.setSerializationInclusion(JsonInclude.Include.NON_EMPTY); 
```
Look at issue: https://github.com/killbilling/recurly-java-library/issues/90

PR #130, opened 2016-07-26
Jackson 2.6 serialization fix

nhartner:
Using Jackson 2.6.X or 2.7.X, I ran into serialization issue when sending a SubscriptionUpdate with multiple SubcriptionAddOn elements. I narrowed this down to an issue between Jackson 2.5 and 2.6+. If you update jackson version to 2.6+ in recurly-java-library, the following unit test fails:
`testDeserializationWithAddons(com.ning.billing.recurly.model.TestSubscription): Can not start an object, expecting field name`

The use of JsonInclude.NON_EMPTY also causes problems in Jackson 2.6, namely that Integer fields with value 0 are treated as null. This caused TestAdjustment unit test to fail with error:
```
java.lang.NullPointerException
	at com.ning.billing.recurly.model.TestAdjustment.testSerialization(TestAdjustment.java:75)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
...
```
JsonInclude.NON_NULL is typically what I usually use and seemed to play better with Jackson 2.6+.

With this patch, all tests pass using Jackson 2.5.X, 2.6.0-2.6.4, and 2.7.4+. There seemed to be another issue using Jackson 2.6.5-2.7.3 with or without my fix.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/RecurlyObject.java
+++ b/src/main/java/com/ning/billing/recurly/model/RecurlyObject.java
@@ -76,7 +76,7 @@ public abstract class RecurlyObject {
         xmlMapper.setAnnotationIntrospector(pair);
         xmlMapper.registerModule(new JodaModule());
         xmlMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
-        xmlMapper.setSerializationInclusion(JsonInclude.Include.NON_EMPTY);
+        xmlMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
 
         final SimpleModule m = new SimpleModule("module", new Version(1, 0, 0, null, null, null));
         m.addSerializer(Accounts.class, new RecurlyObjectsSerializer<Accounts, Account>(Accounts.class, "account"));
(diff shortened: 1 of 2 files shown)


pierre:
Thanks for running the tests against various versions!

We've had other incompatibilities in the past with different Jackson versions:

* https://github.com/killbilling/recurly-java-library/pull/20 (2.1.2 vs 2.2.2)
* https://github.com/killbilling/recurly-java-library/pull/91 (2.3.0 vs 2.5.0)

Maybe we should shade the jars moving forward?

@nhartner: what's the other issue with Jackson 2.6.5-2.7.3? Did you submit a bug report to @cowtowncoder?

@bhelx: do you want to run the enterprise integration tests before merging?

nhartner:
@pierre I didn't submit a report on the other issue. With Jackson it's always a chance they broke something since the issue went away again in 2.7.4+.  But I can file an issue if you like

cowtowncoder:
@nhartner I suspect this fix (for 2.7.0) is relevant here:

https://github.com/FasterXML/jackson-databind/issues/849

where definition of `NON_EMPTY` is different in 2.6; not because of explicit change, but due to earlier missing handling in 2.5, and addition of handling in 2.6. With all the feedback decrying behavior of 2.6, behavior in 2.7 (and going forward in future) was effectively restored to that of 2.5.
So going forward definition is stable and defined.

There may be other parts to it, but I thought it makes sense to clarify this particular behavior since it's one of relatively rare cases of functionality changing and reverting for a minor release. And very unfortunate one at that.

bhelx:
:+1: This passes the integration tests

There is one failing test but it should be taken care of by this closed PR:

https://github.com/killbilling/recurly-java-library/pull/132/files#diff-c98785e3cbe15c609832a056c59ab237L1131

pierre:
Thanks!

PR #131, opened 2016-07-27
Fix po number

jaypatel11:
-- Change "po_number" in Invoice class from Integer to String. "po_number" must be a String per https://dev.recurly.com/docs/post-an-invoice-invoice-pending-charges-on-an-acco
-- Update tests for deserialization


pierre:
👍

bhelx:
@jaypatel11 thanks for catching this. The integration tests pass locally :+1:

PR #132, opened 2016-07-27
Fixes inconsistent integration tests

bhelx:
* Was using a static coupon code value, if the tests failed to clean
  this up then all tests would fail after. Changed it to be random.
* Removed calls to clear billing info. Closing the account suffices.


src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -1128,7 +1110,6 @@ public void testCreateTrialExtensionCoupon() throws Exception {
         final Coupon couponData = TestUtils.createRandomCoupon();
 
         couponData.setName("apitrialext");
-        couponData.setCouponCode("apitrialext");
```
bhelx: This value is already random from `TestUtils.createRandomCoupon();`

pierre:
👍

PR #135, opened 2016-08-02
Abstract pagination methods and http client overriding

MtnWithered:
* adds abstract pagination methods - for generic usage
* allows http client overriding
* adds URI scheme argument in client constructor

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -23,7 +23,6 @@ import java.io.Reader;
 import java.math.BigDecimal;
 import java.net.URL;
 import java.nio.charset.Charset;
-import java.security.GeneralSecurityException;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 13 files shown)


bhelx:
@MtnWithered I think I understand the use case for http client injection, but what would be the use cases for abstracting the pagination methods? Could you go into some more detail is your description?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -160,6 +163,13 @@ public synchronized void open() throws NoSuchAlgorithmException, KeyManagementEx
     }
 
     /**
+     * Use custom http client
+     */
+    public void open(AsyncHttpClient asyncHttpClient) {
```
pierre: My only concern with this is that we expose the implementation to the public API - what if we switch client one day?

How about making `createHttpClient` protected?

MtnWithered:
@bhelx 
I need a generic way to handle the paging in all recurly objects. It's a common feature of all these objects, so an abstract is a natural way of expressing this kind of relationship. I don't want to repeat the code for every RecurlyObjects in my client, so by using an abstraction, I can do it within a single method.
Client overriding is for reconfiguration and testing purposes.

@pierre 
Sure, it's a good idea.

MtnWithered:
Any other suggestions here? Can we proceed with merging?

bhelx:
@MtnWithered The code looks good to me.
@pierre any other concerns?

pierre:
👍

MtnWithered:
Thank you.

PR #142, opened 2016-08-30
New API version 2.4 pagination params

bhelx:
This PR enables using all the query params allowed in API version 2.4

See:
https://dev.recurly.com/docs/pagination
https://dev.recurly.com/docs/list-accounts
https://dev.recurly.com/docs/list-subscriptions
https://dev.recurly.com/docs/list-transactions

### DIFF
--- a/src/main/java/com/ning/billing/recurly/QueryParams.java
+++ b/src/main/java/com/ning/billing/recurly/QueryParams.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2010-2014 Ning, Inc.
+ * Copyright 2014-2015 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.ning.billing.recurly;
+
+import com.ning.billing.recurly.model.Subscription;
+import com.ning.billing.recurly.model.Transaction;
+import org.joda.time.DateTime;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This class is responsible for handling query parameters to pageable resources in the Recurly API.
+ * See the pagination docs (https://dev.recurly.com/docs/pagination) for the parameters available on
+ * every endpoint.
+ */
+public class QueryParams {
+    private static final String RECURLY_PAGE_SIZE_KEY = "recurly.page.size";
+    private static final Integer DEFAULT_PAGE_SIZE = 20;
+
+    public enum Sort {
+        CREATED_AT("created_at"),
+        UPDATED_AT("updated_at");
+
+        private final String type;
+
+        private Sort(final String type) {
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+    }
+
+    public enum Order {
+        ASC("asc"),
+        DESC("desc");
+
+        private final String type;
+
+        private Order(final String type) {
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+    }
+
+    private Map<String,String> params;
+
+    public QueryParams() {
+        params = new HashMap<String, String>();
+
+        Integer pageSize;
+
+        try {
+            pageSize = new Integer(System.getProperty(RECURLY_PAGE_SIZE_KEY));
+        } catch (NumberFormatException nfex) {
+            pageSize = DEFAULT_PAGE_SIZE;
+        }
+
+        setPerPage(pageSize);
+    }
+
+    public void setPerPage(final int perPage) {
+        params.put("per_page", Integer.toString(perPage));
+    }
+
+    public void setSort(final Sort sortField) {
+        params.put("sort", sortField.getType());
+    }
+
+    public void setOrder(final Order orderDirection) {
+        params.put("order", orderDirection.getType());
+    }
+
+    public void setBeginTime(final DateTime beginTime) {
+        params.put("begin_time", formatDate(beginTime));
+    }
+
+    public void setEndTime(final DateTime endTime) {
+        params.put("end_time", formatDate(endTime));
+    }
+
+    private String formatDate(final DateTime dateTime) {
+        return dateTime.toDateTimeISO().toString();
+    }
+
+    public void put(final String key, final String value) { params.put(key, value); }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+
+        for (Map.Entry<String,String> entry : params.entrySet()) {
+            if (sb.length() > 0) {
+                sb.append("&");
+            } else {
+                sb.append("?");
+            }
+
+            sb.append(String.format("%s=%s", entry.getKey(), entry.getValue()));
+        }
+
+        return sb.toString();
+    }
+}
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -17,7 +17,6 @@
 
 package com.ning.billing.recurly;
 
-import java.awt.font.TransformAttribute;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
@@ -50,6 +49,7 @@ import com.ning.billing.recurly.model.Errors;
 import com.ning.billing.recurly.model.GiftCard;
 import com.ning.billing.recurly.model.GiftCards;
 import com.ning.billing.recurly.model.Invoice;
+import com.ning.billing.recurly.model.InvoiceState;
 import com.ning.billing.recurly.model.Invoices;
 import com.ning.billing.recurly.model.Plan;
 import com.ning.billing.recurly.model.Plans;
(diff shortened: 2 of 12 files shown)


bhelx:
@pierre Would love to get your opinion before I move forward. I'd like to deprecate the other API methods that take specific params as arguments in favor of using `QueryParams` or something like it.

src/main/java/com/ning/billing/recurly/QueryParams.java
```diff
@@ -0,0 +1,183 @@
...
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+    }
+
+    // see Subscription pagination: https://dev.recurly.com/docs/list-subscriptions
+    public enum SubscriptionState {
```
bhelx: This could eventually be shared with `Subscription` model

src/main/java/com/ning/billing/recurly/QueryParams.java
```diff
@@ -0,0 +1,183 @@
...
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+    }
+
+    // see Transaction pagination: https://dev.recurly.com/docs/list-transactions
+    public enum TransactionState {
```
bhelx: This could eventually be shared with `Transaction` model

src/main/java/com/ning/billing/recurly/QueryParams.java
```diff
@@ -0,0 +1,183 @@
...
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+    }
+
+    // see Transaction pagination: https://dev.recurly.com/docs/list-transactions
+    public enum TransactionType {
```
bhelx: This could eventually be shared with `Transaction` model

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -440,14 +450,33 @@ public Subscriptions getAccountSubscriptions(final String accountCode) {
...
      * <p>
      * Returns information about a single account.
      *
      * @param accountCode recurly account id
      * @param status      Only accounts in this status will be returned
-     * @return Subscriptions for the specified user
+     * @return Subscriptions on the account
      */
+    @Deprecated
     public Subscriptions getAccountSubscriptions(final String accountCode, final String status) {
```
bhelx: I'd like to deprecate these types of methods in favor of using `QueryParams` or something like it.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -525,6 +554,45 @@ public Transactions getAccountTransactions(final String accountCode) {
...
+    /**
+     * Get site's transaction history
+     * <p>
+     * All transactions on the site
+     *
+     * @param params {@link QueryParams}
+     * @return the transaction history of the site on success, null otherwise
+     */
+    public Transactions getTransactions(final QueryParams params) {
+        final String pathAndQuery = Transactions.TRANSACTIONS_RESOURCE + params.toString();
```
bhelx: This should just be passing the params to `doGET` as it's supported now. Was experimenting here.

pierre:
> @pierre Would love to get your opinion before I move forward.

Ack, I'll take a look (FYI, in general, I won't take a look at a `WIP` PR, unless pinged explicitly).

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -33,6 +33,7 @@
 import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
+import javax.management.Query;
```
pierre: Extra import?

src/main/java/com/ning/billing/recurly/QueryParams.java
pierre: Yeah - these inner enums are weird on a global level. It would make more sense to have them outside or in the associated models.

src/main/java/com/ning/billing/recurly/QueryParams.java
```diff
@@ -0,0 +1,183 @@
...
+
+    public void setEndTime(DateTime endTime) {
+        params.put("end_time", formatDate(endTime));
+    }
+
+    public void setSubscriptionState(SubscriptionState state) {
+        params.put("state", state.getType());
+    }
+
+    public void setTranasctionState(TransactionState state) {
```
pierre: Typo.

src/main/java/com/ning/billing/recurly/QueryParams.java
```diff
@@ -0,0 +1,183 @@
...
+        StringBuilder sb = new StringBuilder();
+
+        for (Map.Entry<String,String> entry : params.entrySet()) {
+            if (sb.length() > 0) {
+                sb.append("&");
+            } else {
+                sb.append("?");
+            }
+
+            sb.append(String.format("%s=%s", entry.getKey(), entry.getValue()));
```
pierre: Good enough for now, but this might need to be url-encoded one day. Could we instead pass the query params map to the async-http-client builder, which will take care of it?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
pierre: I think it would make sense to have something like this:

```
Subscriptions getAccountSubscriptions(final String accountCode, final TransactionStatus status, final QueryParams queryParams)
```

but not everything in `QueryParams` - it's a bit weird to me (should then `accountCode` be in `QueryParams`?). The caller doesn't care if a parameter is sent in the body or in the query parameter.

We actually did something very similar recently in the [killbill-client-java](https://github.com/killbill/killbill-client-java/pull/36).

I would suggest to do one big change in one release, instead of spreading it over several releases, i.e. add new methods for all endpoints, add `@Deprecated` on old methods and do a release. Then we will need to decide how long we want to keep the `@Deprecated` ones.

bhelx: I was thinking more along the lines of:

```
getAccountSubscriptions(final String accountCode, final QueryParams queryParams)
```

Where `status` could be put into the QueryParams. QueryParams are meant to be optional GET params, whereas accountCode is a required part of the url.

pierre: No strong objection if you want to go that way, but here are my thoughts:

* Caller doesn't care about the implementation (what goes into the body, what's optional, etc.), only the functionality, hence I would instead have signatures reflect the functionality rather than the implementation.
* `QueryParams` should be the intersection of all optional fields across APIs, not the union (if only a couple of endpoints accept `TransactionStatus`, I would not put it in there).
* Maybe the name could be more generic (e.g. `OptionalParams`)?

bhelx: > QueryParams should be the intersection of all optional fields across APIs, not the union (if only a couple of endpoints accept TransactionStatus, I would not put it in there).

Initially I was thinking we could have a `QueryParams` class will all the common params then inherit some params with specialized fields. This seemed like overkill and I couldn't think of any other solutions so I'm in agreement with you here.

> Maybe the name could be more generic (e.g. OptionalParams)?

I'm going to think about it. I feel like QueryParams is still a pretty generic name since the spec calls the url part a "Query String"

pierre: > I feel like QueryParams is still a pretty generic name since the spec calls the url part a "Query String"

Sure - my point was simply that it's an implementation detail the consumer of the API doesn't care about, so it's a bit weird to put it in the public API. IMHO, it's similar to asking to specify an `authorizationHeader` instead of simply an `apiKey`/`apiSecret`.

bhelx:
Okay, I've got this updated and got the tests passing. But working on this, I've realized some weird stuff. Mainly that our use of enums is pretty inconsistent. Some of this is my fault, but warrants discussion.

In some places, we have the enum in it's own file:

https://github.com/killbilling/recurly-java-library/blob/6f77b2ece75bbc8090080cfbc28bc8d9fe0409bf/src/main/java/com/ning/billing/recurly/model/RefundOption.java

In another case, we have an enum in the Resource(s) class. Example, Adjustment(s):

https://github.com/killbilling/recurly-java-library/blob/master/src/main/java/com/ning/billing/recurly/model/Adjustments.java#L41

When adding these new enums, I chose the latter option. However it's kind of a problem that they are inconsistent and that might need to be addressed at some point.

Another issue that has now come up is that these new enums aren't used in the resources themselves. For instance, Subscription#getState is still returning a string:

https://github.com/killbilling/recurly-java-library/blob/master/src/main/java/com/ning/billing/recurly/model/Subscription.java#L163

I don't know if we want to add support for these new enums into the models and deprecate the string versions?  

Some more thought needs to go into this i think.

pierre:
> When adding these new enums, I chose the latter option. However it's kind of a problem that they are inconsistent and that might need to be addressed at some point.

Yeah, I had noticed this. In general, having them in their own file can make things easier for ser-des.

> I don't know if we want to add support for these new enums into the models and deprecate the string versions?

Sure, it would be cleaner to return enums as well. But this could be done in a separate PR to minimize changes?

Detailed upgrade notes will be required... there will be a lot of changes! 😼

bhelx:
@pierre I've moved the enums into their own files. I'd like to remove the queryparam builder and pass it on to the underlying library, but I'd prefer to do that in a follow up at this moment.

bhelx:
> Sure, it would be cleaner to return enums as well. But this could be done in a separate PR to minimize changes?
> Detailed upgrade notes will be required... there will be a lot of changes! 😼 

It would be great to make all these API changes at once for a 1.0 release!

bhelx:
@pierre have you had a chance to look at this PR again?

pierre:
Nope and it has been a while... 😼

Was there any other change since we last spoke about it? If not, I think I already went through all of it. This will probably be a major release - will you write some upgrade notes?

bhelx:
@pierre yes, I updated as per your comments

bhelx:
Hey @pierre

Been having a lot of people ask me about this one. Let me know if there is anything else you think I could do to it.

pierre:
> Let me know if there is anything else you think I could do to it.

We can iterate on subsequent PRs, but next steps should be to:

* Prepare some release notes (quite a few changes!)
* Bump major - this could become 1.0.0 if we think the API is stable enough for all use cases?

bhelx:
@pierre sure thing. Thanks for merging!

> Prepare some release notes (quite a few changes!)

It would be great if we kept some changelogs and release notes, but I think for now I will just add some kind of `Pagination` section to the README. I can do that today. Does that sound good?

> Bump major - this could become 1.0.0 if we think the API is stable enough for all use cases?

We should probably do a minor for now. The API is definitely stable but I'd like to get a roadmap of stuff ready for a 1.0.0 release. I can open up an issue and start a wish list for that. I'll reach out to some customers who use the library for feedback.

pierre:
> It would be great if we kept some changelogs and release notes, but I think for now I will just add some kind of `Pagination` section to the README. I can do that today. Does that sound good?

Sure, 👍 !

> We should probably do a minor for now. The API is definitely stable but I'd like to get a roadmap of stuff ready for a 1.0.0 release. I can open up an issue and start a wish list for that. I'll reach out to some customers who use the library for feedback.

💯  for a wish list and customer feedback.

PR #144, opened 2016-09-14
Stabilize integration tests

bhelx:
These tests fail sometimes if the times are slightly off. Since the intention is to test that the datetimes are properly serialized and parseable, I've removed them. This is already tested in the unit tests.


pierre:
👍

PR #146, opened 2016-09-26
Run unit tests only for public PRs

bhelx:
Have travis only run unit tests if the `API_KEY` is not set. `API_KEY` is not set when running tests on public PRs. I have to run the integration tests for every PR anyway.


pierre:
Good idea - looks like the build failed on Travis though.

bhelx:
I think this should be fixed now

pierre:
👍

Issue #143, opened 2016-08-31
Invoice model missing 'closedAt' datetime property


peterjkirby:
According to the Recurly documentation [here](https://dev.recurly.com/docs/list-invoices), an invoice should have a datetime property called `closed_at`. 

Unless I am mistakenly looking in the wrong place, I do not think that the current Invoice implementation has such a property.

bhelx:
I don't see it either. Would be happy to review a PR to add it or I can add it if you don't have the time.

bhelx:
@peterjkirby I went ahead and added this

bhelx:
Closing with #149

PR #149, opened 2016-10-05
Adds closedAt to Invoice

bhelx:
Addresses issue #143

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/Invoice.java
+++ b/src/main/java/com/ning/billing/recurly/model/Invoice.java
@@ -77,6 +77,9 @@ public class Invoice extends RecurlyObject {
     @XmlElement(name = "updated_at")
     private DateTime updatedAt;
 
+    @XmlElement(name = "closed_at")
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/model/Invoice.java
```diff
@@ -307,6 +319,9 @@ public boolean equals(final Object o) {
         if (originalInvoice != null ? !originalInvoice.equals(invoice.originalInvoice) : invoice.originalInvoice != null) {
             return false;
         }
+        if (closedAt != null ? !closedAt.equals(invoice.closedAt) : invoice.closedAt != null) {
```
pierre: Should be instead `closedAt.compareTo(invoice.closedAt) != 0` (FYI it looks like we already have the issue with `createdAt` below).

bhelx: Good catch

pierre:
👍

Issue #123, opened 2016-07-12
401 response from Recurly throws vague exception


pavellishin:
When receiving a 401 response, Recurly doesn't send any XML data, but a simple string: 

`HTTP Basic: Access denied.`

This causes the following exception:

```
Exception in thread "main" com.ning.billing.recurly.RecurlyAPIException: Unspecified API Error
	at com.ning.billing.recurly.RecurlyClient.callRecurly(RecurlyClient.java:1004)
	at com.ning.billing.recurly.RecurlyClient.callRecurlySafe(RecurlyClient.java:941)
	at com.ning.billing.recurly.RecurlyClient.doGETWithFullURL(RecurlyClient.java:895)
	at com.ning.billing.recurly.RecurlyClient.doGET(RecurlyClient.java:888)
	at com.ning.billing.recurly.RecurlyClient.getPlan(RecurlyClient.java:629)
	at com.bazaarvoice.connections.greenlight.api.Recurly.main(Recurly.java:19)
13:38:41.479 [main] DEBUG c.ning.billing.recurly.RecurlyClient - Unable to extract error
java.io.IOException: javax.xml.stream.XMLStreamException: ParseError at [row,col]:[1,1]
Message: Content is not allowed in prolog.
	at com.fasterxml.jackson.dataformat.xml.util.StaxUtil.throwXmlAsIOException(StaxUtil.java:24)
	at com.fasterxml.jackson.dataformat.xml.XmlFactory._createParser(XmlFactory.java:557)
	at com.fasterxml.jackson.dataformat.xml.XmlFactory.createParser(XmlFactory.java:429)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2779)
	at com.ning.billing.recurly.RecurlyClient.callRecurly(RecurlyClient.java:1000)
	at com.ning.billing.recurly.RecurlyClient.callRecurlySafe(RecurlyClient.java:941)
	at com.ning.billing.recurly.RecurlyClient.doGETWithFullURL(RecurlyClient.java:895)
	at com.ning.billing.recurly.RecurlyClient.doGET(RecurlyClient.java:888)
	at com.ning.billing.recurly.RecurlyClient.getPlan(RecurlyClient.java:629)
	at com.bazaarvoice.connections.greenlight.api.Recurly.main(Recurly.java:19)
Caused by: javax.xml.stream.XMLStreamException: ParseError at [row,col]:[1,1]
[...]

PR #151, opened 2016-10-05
Throws a less vague exception for unauth calls

bhelx:
Addresses #123

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -1072,6 +1072,7 @@ public class RecurlyClient {
         if (response.getStatusCode() != 200) {
             RecurlyAPIError recurlyAPIError = new RecurlyAPIError();
             recurlyAPIError.setHttpStatusCode(response.getStatusCode());
+
             throw new RecurlyAPIException(recurlyAPIError);
         }
 
@@ -1162,6 +1163,7 @@ public class RecurlyClient {
             // Handle errors payload
             if (response.getStatusCode() >= 300) {
                 log.warn("Recurly error whilst calling: {}\n{}", response.getUri(), payload);
+                RecurlyAPIError recurlyError = new RecurlyAPIError();
 
                 if (response.getStatusCode() == 422) {
                     final Errors errors;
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -1176,11 +1177,18 @@ private void doDELETE(final String resource) {
...
-                        log.debug("Unable to extract error", e);
+
+                    if (response.getStatusCode() == 401) {
+                        recurlyError.setSymbol("unauthorized");
+                        recurlyError.setDescription("You don't have access to this resource. Either your subdomain and private key are incorrect or this resource is private.");
+                    } else {
+                        try {
+                            recurlyError = xmlMapper.readValue(payload, RecurlyAPIError.class);
+                        } catch (Exception e) {
+                            log.debug("Unable to extract error", e);
```
pavellishin: Would help to throw the contents of `payload` into the logs as well.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -1176,11 +1177,18 @@ private void doDELETE(final String resource) {
                     throw new TransactionErrorException(errors);
                 } else {
                     RecurlyAPIError recurlyError = new RecurlyAPIError();
-                    try {
-                        recurlyError = xmlMapper.readValue(payload, RecurlyAPIError.class);
-                    } catch (Exception e) {
-                        log.debug("Unable to extract error", e);
+
+                    if (response.getStatusCode() == 401) {
```
pavellishin: Why not make this an `else if` with the error above? 

```
if (response.getStatusCode() == 422) {
  // 422 handling stuff here
} else if (response.getStatusCode() == 401) {
  // update recurlyError as you're doing here
} else {
 // try/catch the xmlMapper.readValue block
}
```

pavellishin:
I haven't checked; does a 403 behave like a 401, with a non-xml response?

Also, I guess the rest - for some reason, Recurly only shows examples for the 404 and 422 cases here: https://dev.recurly.com/docs/welcome

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: Yeah, looks less nested

bhelx:
@pavellishin I can't think of any scenario where our server throws 403 for API v2. The reason that the docs show the 404 and 422 cases are because those are the error codes which contain descriptive XML and those are the codes that you may need to parse and inform your user to fix (if you have a user on the other end). I think it could be more clear what's going on in the docs so I'll make a task to check it out.

Issue #152, opened 2016-11-02
revenue_schedule_type Null Pointer Exception


jfreeby:
We just recently upgraded to 0.6.2 and are getting exceptions when looking up some subscriptions:

com.fasterxml.jackson.databind.JsonMappingException: N/A (through reference chain: com.ning.billing.recurly.model.Subscriptions["subscription"]->com.ning.billing.recurly.model.Subscription["revenue_schedule_type"])

Caused by: java.lang.NullPointerException
	at com.ning.billing.recurly.model.Subscription.setRevenueScheduleType(Subscription.java:338)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:97)
	... 45 more

When I perform a curl request for the subscription that fails, Recurly is returning an empty element:

 <revenue_schedule_type></revenue_schedule_type>

For other requests that succeed there is a value in the element

 <revenue_schedule_type>evenly</revenue_schedule_type>

Please let me know if you need additional information.

Thanks,

Jim

jfreeby:
More info .....

The null pointer is happening in the method, where the revenueScheduleType is null and the call to toUpperCase fails.

public void setRevenueScheduleType(final String revenueScheduleType) {
    this.revenueScheduleType = RevenueScheduleType.valueOf(revenueScheduleType.toUpperCase());
}

bhelx:
Thanks @jfreeby

I'll take a look at this and try to reproduce it

jfreeby:
A simple null check fixed our issue.  Do you have any idea when this might make it to a release?  I have a custom version for now.

bhelx:
@jfreeby if you send a PR, it should be a short review.

jfreeby:
@bhelx Can you point to any instructions on sending a PR?  I just checked the code out with SVN and made the changes locally.  I'd be happy to send the code in any way possible.

bhelx:
@jfreeby You'll want to follow the standard Github procedure of forking the repository, creating a branch, and sending the PR into the master branch of this repo: https://help.github.com/articles/creating-a-pull-request-from-a-fork/

[...]

PR #153, opened 2016-11-03
Added null check for revenueScheduleType

jfreeby:
Resolves: https://github.com/killbilling/recurly-java-library/issues/152

Added a null check for the String revenueScheduleType.

You may want to consider updating the enum RevenueScheduleType to return a default value if the string is empty

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/Subscription.java
+++ b/src/main/java/com/ning/billing/recurly/model/Subscription.java
@@ -341,7 +341,9 @@ public class Subscription extends AbstractSubscription {
     }
 
     public void setRevenueScheduleType(final String revenueScheduleType) {
-        this.revenueScheduleType = RevenueScheduleType.valueOf(revenueScheduleType.toUpperCase());
+    	if( revenueScheduleType != null ){
+    		this.revenueScheduleType = RevenueScheduleType.valueOf(revenueScheduleType.toUpperCase());
+    	}
     }
 
     public GiftCard getGiftCard() {
--- a/src/test/java/com/ning/billing/recurly/model/TestSubscription.java
+++ b/src/test/java/com/ning/billing/recurly/model/TestSubscription.java
@@ -177,7 +177,7 @@ public class TestSubscription extends TestModelBase {
     	try{
         	//we expect an exception here
         	subscription.setRevenueScheduleType( invalidValue );
-        	assertEquals(true, false );
+        	Assert.fail();
         }
         catch( IllegalArgumentException iae ){
         	//iae.printStackTrace();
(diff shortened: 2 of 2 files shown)


src/main/java/com/ning/billing/recurly/model/Subscription.java
```diff
@@ -341,7 +341,9 @@ public RevenueScheduleType getRevenueScheduleType() {
     }
 
     public void setRevenueScheduleType(final String revenueScheduleType) {
-        this.revenueScheduleType = RevenueScheduleType.valueOf(revenueScheduleType.toUpperCase());
+    	if( revenueScheduleType != null ){
```
pierre: Could you add a (fast, i.e. unit) regression test?

jfreeby: @pierre I can added a unit test for a null value and for valid values.

How do you want to handle the case where an invalid string is added?  One where there is no matching RevenueScheduleType

For example 
RevenueScheduleType.valueOf("INVALID_STRING") will return an exception

java.lang.IllegalArgumentException: No enum constant com.ning.billing.recurly.model.RevenueScheduleType.INVALID_STRING

At this point, an exception is thrown

src/test/java/com/ning/billing/recurly/model/TestSubscription.java
```diff
@@ -156,6 +156,40 @@ public void testHashCodeAndEquality() throws Exception {
...
+        	verifyRevenueScheduleType( revenueScheduleType );
+        }
+    }
+    
+    private void verifyRevenueScheduleTypeWithInvalidValue( String invalidValue ){
+    	Subscription subscription = new Subscription();
+    	try{
+        	//we expect an exception here
+        	subscription.setRevenueScheduleType( invalidValue );
+        	assertEquals(true, false );
```
pierre: Nit: `assertFail()` instead.

jfreeby:
I updated with Assert.fail()

jfreeby:
Is it possible that Recurly we might get an empty string from Recurly?  If yes, then we may need a check for an empty string.  Thoughts?

bhelx:
@pierre @jfreeby `IllegalArgumentException` cool with me

bhelx:
> Is it possible that we might get an empty string from Recurly? If yes, then we may need a check for an empty string. Thoughts?

@jfreeby As your code stands, having an empty string in the xml `<revenue_schedule_type></revenue_schedule_type>` gives us a null which is what i would expect. I don't think it should return a default as it is a nullable field.

```java
Assert.assertEquals(subscription.getRevenueScheduleType(), null);
```

There is a separate problem though. I dug into the server side code and realized that this null is not being rendered properly for this field. It should be rendering:

`<revenue_schedule_type nil="nil"></revenue_schedule_type>`

But I can deal with that myself in another PR. It's a little more dangerous changing the servers responses.

bhelx:
This PR seems good to me. Any other comments @pierre ?

pierre:
Nit: style, but otherwise :shipit:!

bhelx:
Okay, I'm going to do a follow up PR to fix formatting.

PR #157, opened 2016-11-11
List site subscriptions

bhelx:
* Removes an unused import 
* Adds the site level `/subscriptions` endpoint


pierre:
Simple enough - maybe a test eventually? 😼

bhelx:
Yeah, we should probably have a integration test for every method in `RecurlyClient`:
https://github.com/killbilling/recurly-java-library/pull/158

PR #158, opened 2016-11-14
Add integration test for getSubscriptions()

bhelx:
Adds a test for #157


bhelx:
I had to remove this broken part of the transaction test that was trying to refund a transaction. I think i'll need to add another test around refunds.

bhelx:
This is ready to go

pierre:
🚢  🇮🇹

PR #159, opened 2016-12-13
Make travis badge reflect master branch only

bhelx:
The travis badge was just showing the latest build (which would include
failing experiemental branches). We want it to only show master which is
green and makes people feel good when they see the README :)


pierre:
:shipit:

PR #160, opened 2016-12-13
Vertex Integration

bhelx:
This is going into the API version 2.5 branch which is being held off until 2.5 is stable. These are some read only attributes and it only applies to sites with a special Vertex tax integration (which will only be one starting at the turn of the year).

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/JurisDetail.java
+++ b/src/main/java/com/ning/billing/recurly/model/JurisDetail.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2010-2014 Ning, Inc.
+ * Copyright 2014-2015 The Billing Project, LLC
+ *
+ * The Billing Project licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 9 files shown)


pierre:
Do you already know how different the integration will look like compared to `Avalara`? Will we have vendor specific APIs or generic tax APIs?

bhelx:
@pierre We are trying to make it as generic as possible, but it's unclear right now whether or not this will become a full integration that we offer to other customers.

bhelx:
I'm adding one integration test for this then it should be good to go.

bhelx:
Okay, I have added an integration test to make sure you get a "feature_not_enabled" error if you try to use the vertex fields.

src/main/java/com/ning/billing/recurly/model/JurisDetail.java
```diff
@@ -0,0 +1,130 @@
...
+        if (jurisdiction != null ? !jurisdiction.equals(jurisDetail.jurisdiction) : jurisDetail.jurisdiction != null) {
+            return false;
+        }
+        if (jurisdictionName != null ? !jurisdictionName.equals(jurisDetail.jurisdictionName) : jurisDetail.jurisdictionName != null) {
+            return false;
+        }
+        if (taxInCents != null ? !taxInCents.equals(jurisDetail.taxInCents) : jurisDetail.taxInCents != null) {
+            return false;
+        }
+        if (rate != null ? !rate.equals(jurisDetail.rate) : jurisDetail.rate != null) {
```
pierre: Need to use `compareTo` for `BigDecimal`.

pierre:
> it's unclear right now whether or not this will become a full integration that we offer to other customers

Are you saying that Avalara is the preferred integration but select customers can use Vertex instead?

bhelx:
Thanks for reviewing! Updated that equals method in question.

> Are you saying that Avalara is the preferred integration but select customers can use Vertex instead?

Yes, everyone automatically has some level of integration with Avalara and we also encourage people to try out the "direct" integration with them. This Vertex feature is currently only for a small subset of enterprise customers who need more details about their taxes. We are still investigating whether or not we will make it a public offering.

PR #167, opened 2017-01-11
Fix integration test by ensuring giftcard.deliver_at always in future

bhelx:
This fixes the integration test. I had unfortunately added a static datetime for deliver_at but that value needs to always be 1 hour in the future. This sets it to 5 days from the day of running.


pierre:
👍

PR #168, opened 2017-01-11
Fix query param generation on getTransactions

### LINKED ISSUE #166: Get All Purchase Transactions fails
<version>0.7.0</version>

When calling recurlyClient.getTransactions(TransactionState.SUCCESSFUL, TransactionType.PURCHASE , queryParams);

The call fails with a message:

<?xml version="1.0" encoding="UTF-8"?>
<error>
  <symbol>type_invalid</symbol>
  <description>Invalid type (must be one of authorization, purchase, refund)</description>
</error>

During execution, the following method is called (with a new QueryParams that has a default page size)

The resource already has some QueryParams added
    private String constructGetUrl(final String resource, QueryParams params) {
        return baseUrl + resource + params.toString();
    }

The resulting URL looks like:

/v2/transactions?per_page=20&end_time=2017-01-16T00:00:00.000-08:00&begin_time=2017-01-04T00:00:00.000-08:00&state=successful&type=purchase?per_page=20 HTTP/1.1

You can see the 2nd "?per_page=20" added to the URL.  This makes the "type" value invalid.

bhelx:
Resolves #166

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -677,9 +677,7 @@ public class RecurlyClient {
         if (state != null) params.put("state", state.getType());
         if (type != null) params.put("type", type.getType());
 
-        final String pathAndQuery = Transactions.TRANSACTIONS_RESOURCE + params.toString();
-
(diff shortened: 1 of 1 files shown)


pierre:
👍

PR #172, opened 2017-01-13
Support open amount invoice refund

### LINKED ISSUE #169: Add support for Invoice Refunds
Version 0.7.0 does not support invoice refunds. We can not use Transaction Refunds because our account was created after November 11, 2014.
According to Recurly docs:`We are moving to refunds through the Invoice object. Transaction refunds will no longer work if you have switched to Invoice Refunds in your Site Settings. If you created your site after November 11, 2014, you are automatically on Invoice refunds. Please see our API documentation on Invoices.`

bhelx:
Resolves #169 

This PR is intended to support both open amount refunds on invoices. It's still a work in progress. I've decided to move the other refund method to a follow up PR.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -794,7 +794,7 @@ public class RecurlyClient {
     }
 
     /**
-     * Refund an invoice given some adjustments
+     * Refund an invoice given an open amount
      * <p/>
      * Returns the refunded invoice
      *
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -49,6 +49,7 @@ import com.ning.billing.recurly.model.Errors;
 import com.ning.billing.recurly.model.GiftCard;
 import com.ning.billing.recurly.model.GiftCards;
 import com.ning.billing.recurly.model.Invoice;
+import com.ning.billing.recurly.model.InvoiceRefund;
 import com.ning.billing.recurly.model.InvoiceState;
 import com.ning.billing.recurly.model.Invoices;
 import com.ning.billing.recurly.model.Plan;
@@ -58,6 +59,7 @@ import com.ning.billing.recurly.model.RecurlyObject;
 import com.ning.billing.recurly.model.RecurlyObjects;
 import com.ning.billing.recurly.model.Redemption;
 import com.ning.billing.recurly.model.Redemptions;
+import com.ning.billing.recurly.model.RefundApplyOrder;
 import com.ning.billing.recurly.model.RefundOption;
 import com.ning.billing.recurly.model.ShippingAddresses;
 import com.ning.billing.recurly.model.Subscription;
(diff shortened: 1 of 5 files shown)


src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -154,6 +157,7 @@ public void testRemoveSubscriptionAddons() throws Exception {
 
             // Create a plan with addons
             final Plan plan = recurlyClient.createPlan(planData);
+            Thread.sleep(1000);
```
pierre: There is no better synchronization option? Could we poll the server instead maybe?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -773,6 +794,24 @@ public Invoices getAccountInvoices(final String accountCode) {
...
+     * Refund an invoice given some adjustments
+     * <p/>
+     * Returns the refunded invoice
+     *
+     * @param invoiceId The id of the invoice to refund
+     * @param amountInCents The open amount to refund
+     * @param order If credit line items exist on the invoice, this parameter specifies which refund method to use first
+     * @return the refunded invoice
+     */
+    public Invoice refundInvoice(final String invoiceId, final Integer amountInCents, final RefundApplyOrder order) {
```
pierre: Can an invoice have several (partial) payments?

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
bhelx: Yes, not sure why this started happening all of the sudden but I am investigating why and trying to determine a better solution.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: You can only refund an invoice that was successfully paid. This method allows you to refund an open amount on the invoice. Since the invoice may contain transactions and credits, it allows you to define which to prioritize in the refund. We have another method that allows you to pass in specific transaction refund objects but I've decided to add that one in a follow up.

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
pierre: 👍

bhelx: Okay, I figured out this is a temporary thing on the server side. It should be operating normally by next week. I'm going to add a TODO tag here and get rid of this next week.

bhelx:
okay @pierre 

Removed WIP and hopefully addressed some of your comments. I'll have the follow up for the other refund method out this week and I can remove those `Thread.sleep()` next week.

pierre:
> I'll have the follow up for the other refund method out this week and I can remove those `Thread.sleep()` next week.

👍 Once we have these, I'll merge and trigger a release.

pierre:
@bhelx Any update on the `sleep`?

cfrasineanu-intx:
Our project will need this change soon. Do you have the timeline? Thank you

bhelx:
@pierre I was told that the change needed to remove the sleep was pushed back two weeks. I think it's probably best to leave the sleeps then I can come back with a follow up in two weeks.

pierre:
Aww 😿 

Alright, I'll merge it as-is to unblock @cfrasineanu-intx.

pierre:
@cfrasineanu-intx 0.8.0 is on its way to Maven Central.

bhelx:
Thanks @pierre!

PR #173, opened 2017-01-19
Replace uses of Guava's Objects#firstNonNull with MoreObjects#firstNonNull 

### LINKED ISSUE #171: Replace uses of Guava's Objects#firstNonNull with MoreObjects#firstNonNull
I tried to use Guava v.21 in our application and got a runtime error because RecurlyClient.java uses  com.google.common.base.Objects#firstNonNull (which was removed Guava 21). This method was deprecated in v.18 and it is recommended to use com.google.common.base.MoreObjects#firstNonNull instead. Thank you!

bhelx:
Resolves #171

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -82,7 +82,7 @@ import com.ning.http.client.Response;
 
 import com.fasterxml.jackson.dataformat.xml.XmlMapper;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Objects;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.StandardSystemProperty;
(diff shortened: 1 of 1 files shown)


bhelx:
@pierre are there any implications to this for people who may be using an older version?

cfrasineanu-intx:
Guava 21 requires Java 8

bhelx:
@cfrasineanu-intx thanks, I did not notice this. Perhaps just switching to using `MoreObjects` is enough for right now.

cfrasineanu-intx:
enough for our project. Thanks

pierre:
> @pierre are there any implications to this for people who may be using an older version?

Upgrading transitive dependencies is always tricky, especially for people using Guava in their project. We don't really know if they rely on a specific version, which could be incompatible with the version we need.

👍 for non upgrading, unless we have to.

PR #175, opened 2017-01-25
Support line item refunds

bhelx:
This adds line item invoice refunds as discussed in #169 . https://dev.recurly.com/docs/line-item-refunds

You have to pass in a list of `AdjustmentRefund`s because they have some fields that don't really belong in `Adjustment`. I've created a convenience method for turning an `Adjustment` into an `AdjustmentRefund`.  You can see the integration test to get an idea how to use it.

```java
final invoice = recurlyClient.getInvoice("1234");
final ArrayList<AdjustmentRefund> lineItems = new ArrayList<AdjustmentRefund>();

// let's just refund the first adjustment
// we can use "toAdjustmentRefund" on the adjustment to turn it
// into an AdjustmentRefund  
final AdjustmentRefund adjustmentRefund = invoice.getLineItems().get(0).toAdjustmentRefund();

// we could change the quantity here or the prorating settings if we want, defaults to full quantity
// adjustmentRefund.setQuantity(1);
lineItems.add(adjustmentRefund);

final Invoice refundInvoice = recurlyClient.refundInvoice(invoice.getId(), lineItems, RefundApplyOrder.transaction);
```

### DIFF
--- a/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
+++ b/src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
@@ -30,6 +30,7 @@ import com.ning.billing.recurly.model.Accounts;
 import com.ning.billing.recurly.model.AddOn;
 import com.ning.billing.recurly.model.AddOns;
 import com.ning.billing.recurly.model.Adjustment;
+import com.ning.billing.recurly.model.AdjustmentRefund;
 import com.ning.billing.recurly.model.Adjustments;
 import com.ning.billing.recurly.model.BillingInfo;
 import com.ning.billing.recurly.model.Coupon;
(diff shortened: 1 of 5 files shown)


bhelx:
@pierre this is also ready for review. Commits cleaned up.

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -1484,4 +1485,72 @@ public void testInvoiceRefund() throws Exception {
...
+            // into an AdjustmentRefund
+            final AdjustmentRefund adjustmentRefund = invoice.getLineItems().get(0).toAdjustmentRefund();
+
+            // we could change the quantity here or the prorating settings if we want, defaults to full quantity
+            // adjustmentRefund.setQuantity(1);
+            lineItems.add(adjustmentRefund);
+
+            final Invoice refundInvoice = recurlyClient.refundInvoice(invoice.getId(), lineItems, RefundApplyOrder.transaction);
+
+            Assert.assertEquals(refundInvoice.getTotalInCents(), new Integer(-100));
```
pierre: `refundInvoice ` will contain 3 adjustments, 2 charges and 1 credit, correct? In that case, shouldn't it be 100 (total amount charged, i.e. $200 charged - $100 refunded)?

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -1484,4 +1485,72 @@ public void testInvoiceRefund() throws Exception {
...
+            Assert.assertEquals(refundInvoice.getSubtotalInCents(), new Integer(-100));
```
pierre: Out of curiosity, what's `subtotal_in_cents`? I couldn't find it in the docs.

bhelx: It's the total before taxes. So all the adjustments summed.

src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
bhelx: That's not quite the way it works. This invoice has 2 adjustments of $1.00 each. I am refunding 1 of the adjustments. The refund invoice contains one adjustment per `adjustmentRefund` given but a negative amount. So if you refunded both adjustments here you would have 2 line items of -100 + -100 = total -200

bhelx: Sorry I saw this last week but forgot to come back and comment :)

pierre: 👍

bhelx:
Thanks again @pierre!

PR #177, opened 2017-02-14
Add product_code for Vertex integration

bhelx:
This adds `product_code` on Transaction and Adjustment as needed for Vertex integration.


bhelx:
I changed `Transaction#getProductCode` to be protected since it's a write-only attribute.

PR #183, opened 2017-03-31
[V2.6] Cardless Free Trial changes

bhelx:
This is going into the API version 2.6. Some more changes related to cardless free trials.


pierre:
👍

PR #188, opened 2017-05-09
Fix integration tests

bhelx:
This fixes the integration tests. I think something may have changed around remainingBillingCycles but I cannot tell when. I am looking into the change on the server side. If this does appear to be a server side bug I will change the tests back.


pierre:
> If this does appear to be a server side bug I will change the tests back.

👍  I'll let you merge this PR if/when you confirm it.

It would be quite a thing if the client tests found a server-side bug! 😸

bhelx:
Thanks, I'm looking into this today.

> It would be quite a thing if the client tests found a server-side bug! 😸

I agree :) !

PR #189, opened 2017-05-11
Remove X-Records parsing and add count methods

bhelx:
This is a new change in API version 2.6 and is going into the `api_version_2_6` branch.

We no longer return `X-Records` on every page. Instead, if you wish to get the count you must explicitly call the server with a `HEAD` request and parse the `X-Records` header.

This will be a breaking change because we no longer will support `RecurlyObjects#getNbRecords()`.

The integration tests will be healed by this: https://github.com/killbilling/recurly-java-library/pull/188

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/RecurlyObjects.java
+++ b/src/main/java/com/ning/billing/recurly/model/RecurlyObjects.java
@@ -50,8 +50,6 @@ public abstract class RecurlyObjects<T extends RecurlyObject> extends ArrayList<
     @XmlTransient
     private String nextUrl;
 
-    @XmlTransient
-    private Integer nbRecords;
 
     @JsonIgnore
     <U extends RecurlyObjects> U getStart(final Class<U> clazz) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/ning/billing/recurly/model/RecurlyObjects.java
```diff
@@ -98,16 +96,6 @@ public void setNextUrl(final String nextUrl) {
         this.nextUrl = nextUrl;
     }
 
-    @JsonIgnore
-    public Integer getNbRecords() {
```
bhelx: This is the breaking change

pierre: Note: we could have made it backward compatible, by passing a handler (instead of adding new APIs) which would be executed only when the method is explicitly called. But probably not worth the extra complexity.

bhelx: Making a note for 1.0 ideas

Issue #190, opened 2017-05-16
Provide more info in FailedPaymentNotification


cfrasineanu-intx:
Hello, 
In our project we need to use the following data that Recurly returns in the failed_payment_notification webhook:
`<gateway_error_codes nil="true"></gateway_error_codes>
    <failure_type>Declined by the gateway</failure_type>`
I can make the necessary code changes.
Thanks!

bhelx:
@cfrasineanu-intx it's possible our webhooks parsers are a little behind. I'll merge any changes you want to make just tag me on the PR!

bhelx:
Closed with #191

PR #191, opened 2017-05-18
Added failureType and gatewayErrorCodes to PushNotification (issue 190)

cfrasineanu-intx:
Addresses #190

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/push/payment/PushTransaction.java
+++ b/src/main/java/com/ning/billing/recurly/model/push/payment/PushTransaction.java
@@ -103,7 +103,7 @@ public class PushTransaction extends AbstractTransaction {
         return failureType;
     }
 
-    public void setFailureType(Object failureType) {
+    public void setFailureType(final Object failureType) {
         this.failureType = stringOrNull(failureType);
     }
 
@@ -111,7 +111,7 @@ public class PushTransaction extends AbstractTransaction {
         return gatewayErrorCodes;
     }
 
-    public void setGatewayErrorCodes(Object gatewayErrorCodes) {
+    public void setGatewayErrorCodes(final Object gatewayErrorCodes) {
         this.gatewayErrorCodes = stringOrNull(gatewayErrorCodes);
     }
 
--- a/src/main/java/com/ning/billing/recurly/model/push/payment/PushTransaction.java
+++ b/src/main/java/com/ning/billing/recurly/model/push/payment/PushTransaction.java
@@ -45,6 +45,12 @@ public class PushTransaction extends AbstractTransaction {
     @XmlElement
     private String message;
 
+    @XmlElement(name = "failure_type")
+    private String failureType;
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/model/push/payment/PushTransaction.java
```diff
@@ -93,7 +99,21 @@ public void setMessage(final Object message) {
...
+    public void setFailureType(Object failureType) {
+        this.failureType = stringOrNull(failureType);
+    }
+
+    public String getGatewayErrorCodes() {
+        return gatewayErrorCodes;
+    }
 
+    public void setGatewayErrorCodes(Object gatewayErrorCodes) {
+        this.gatewayErrorCodes = stringOrNull(gatewayErrorCodes);
```
pierre: Question for @bhelx: according to the docs, it looks like a [String](https://dev.recurly.com/page/webhooks) indeed, but why the plural? Can one expect multiple codes as CSV? If so, maybe the library could deserialize them right away for convenience.

cfrasineanu-intx: I could not find the XML schema (or DTD) for the Recurly notifications. Do you know if one exists? It would clarify the issue. Thanks

bhelx: @cfrasineanu-intx sorry we do not have an xsd or schema.
@pierre this is indeed a String. I believe the format of this string is dependent on the gateway in which the error comes back from but I am not sure. Either way I think we want to leave it as a String and let the programmer decide what to do with it.

src/main/java/com/ning/billing/recurly/model/push/payment/PushTransaction.java
```diff
@@ -93,7 +99,21 @@ public void setMessage(final Object message) {
         this.message = stringOrNull(message);
     }
 
+    public String getFailureType() {
+        return failureType;
+    }
+
+    public void setFailureType(Object failureType) {
```
bhelx: Can we mark this setter argument as `final`?

src/main/java/com/ning/billing/recurly/model/push/payment/PushTransaction.java
```diff
@@ -93,7 +99,21 @@ public void setMessage(final Object message) {
...
+        this.failureType = stringOrNull(failureType);
+    }
+
+    public String getGatewayErrorCodes() {
+        return gatewayErrorCodes;
+    }
 
+    public void setGatewayErrorCodes(Object gatewayErrorCodes) {
```
bhelx: Can we mark this setter argument as `final`?

bhelx:
Thanks @cfrasineanu-intx !

cfrasineanu-intx:
@bhelx You're welcome!

PR #197, opened 2017-06-01
Add coupon_code to SubscriptionUpdate

lukevs:
Currently, there is no way to add a coupon when updating a subscription. In order to transfer to a subscription with a coupon, users need to terminate the existing subscription and create a new one.

This PR adds the field for the supported coupon_code attribute on subscription updates.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/SubscriptionUpdate.java
+++ b/src/main/java/com/ning/billing/recurly/model/SubscriptionUpdate.java
@@ -54,6 +54,17 @@ public class SubscriptionUpdate extends AbstractSubscription {
         this.collectionMethod = stringOrNull(collectionMethod);
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/model/SubscriptionUpdate.java
```diff
@@ -54,6 +54,17 @@ public void setCollectionMethod(Object collectionMethod) {
...
     }
 
+    @XmlElement(name = "coupon_code")
+    private String couponCode;
+
+    public String getCouponCode() {
+        return couponCode;
+    }
+
+    public void setCouponCode(String couponCode) {
```
bhelx: Normally we mark these as `final`. I will push a follow up for it later.

bhelx:
👍

PR #199, opened 2017-06-16
Add missing coupon code getter

jaypatel11:


bhelx:
@jaypatel11 thank you for catching this!

PR #205, opened 2017-07-27
update guava to 22.0

hijklmno:
Update guava to 22.0. Tracking WIP here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <jackson.version>2.5.0</jackson.version>
-        <guava.version>18.0</guava.version>
+        <guava.version>22.0</guava.version>
         <slf4j.version>1.7.10</slf4j.version>
     </properties>
     <dependencies>


bhelx:
@hijklmno Is any more work needed with this? What are the implications of merging this?

@pierre any opinions on it?

pierre:
> @pierre any opinions on it?

There are some side effects doing the upgrade (release 21.0 requires JDK 1.8 or newer, removal of deprecated methods, etc.).

That being said, I don't have a strong opinion. However, in general, I tend to avoid unnecessary upgrades as it adds some overhead to the user of the library (who is forced to upgrade the third-party library as well), unless it becomes a maintenance nightmare for us (or security/bug fixes).

hijklmno:
Hey @bhelx, thanks for getting back to this. I am currently on a different project and will return to recurly sometime early to mid October. The main motivation for this is initially we were checking in the source code of this library (back at version 0.6.0) into our code base, and have made modifications on it to suit our needs. The current version of guava (18.0) conflicts with our current version (21.0). We are also evaluating artifactory on the side.

Otherwise, I agree with @pierre. I will have a more detailed and researched answer by then.

hijklmno:
Closing due to no longer needed.

PR #209, opened 2017-08-08
Add ability to set subscription UUID when redeeming a coupon

mslipper:
Context:

As per https://dev.recurly.com/docs/redeem-a-coupon-before-or-after-a-subscription, the Recurly API supports the ability to pass in a subscription_uuid as part of the request to redeem a coupon. I noticed that there was no ability to do that through this library, so I implemented that ability. Please let me know if there are any changes you'd like me to make as part of code review and I'd be happy to add them. I've run the test suite locally and everything passes.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/Redemption.java
+++ b/src/main/java/com/ning/billing/recurly/model/Redemption.java
@@ -82,8 +82,8 @@ public class Redemption extends RecurlyObject {
         return subscriptionUuid;
     }
 
-    public void setSubscriptionUuid(String subscriptionUuid) {
-        this.subscriptionUuid = subscriptionUuid;
+    public void setSubscriptionUuid(final Object subscriptionUuid) {
+        this.subscriptionUuid = stringOrNull(subscriptionUuid);
     }
 
     public Coupon getCoupon() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/ning/billing/recurly/model/Redemption.java
```diff
@@ -75,6 +78,14 @@ public void setAccountCode(final Object accountCode) {
         this.accountCode = stringOrNull(accountCode);
     }
 
+    public String getSubscriptionUuid() {
+        return subscriptionUuid;
+    }
+
+    public void setSubscriptionUuid(String subscriptionUuid) {
```
pierre: `Object` should be used instead (see `setAccountCode` above for an example).

bhelx: I agree with @pierre. See how we do it with other String setters. Ex: https://github.com/killbilling/recurly-java-library/pull/209/files#diff-b42990d64ddafc3a2661798e66ee9026R131

Let's also mark it `final`.

mslipper:
@bhelx got it! Will fix in the next few hours. Thanks for the prompt review :)

mslipper:
Updates in!

PR #216, opened 2017-09-27
Custom Invoice Notes, ISO 3166 country codes

bhelx:
**NOTE**: The tests will pass for this on Thursday or Friday. We can hold off until then.

This is for API v2.8 and is only going into the `api_version_2_8` branch for now.

* Added the ability to set custom invoice notes on the `Purchase` object. It seems we missed adding this in `Invoice` in a previous version so I added them there too.
* Updated the tests to use valid 2 character ISO 3166 country codes (which we now validate).
* Fixed a mistake I made in the `TestPurchase` file. We were not properly deserializing and testing the purchase object.

### DIFF
--- a/src/test/java/com/ning/billing/recurly/model/TestPurchase.java
+++ b/src/test/java/com/ning/billing/recurly/model/TestPurchase.java
@@ -34,6 +34,9 @@ public class TestPurchase extends TestModelBase {
                 "  <collection_method>automatic</collection_method>" +
                 "  <net_terms>30</net_terms>" +
                 "  <currency>USD</currency>" +
+                "  <customer_notes>Customer Notes</customer_notes>" +
+                "  <terms_and_conditions>Terms and Conditions</terms_and_conditions>" +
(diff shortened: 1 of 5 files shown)


src/test/java/com/ning/billing/recurly/model/TestPurchase.java
```diff
@@ -124,8 +130,7 @@ public void testSerialization() throws Exception {
         purchase.setCouponCodes(couponCodes);
         purchase.setGiftCard(giftCard);
 
-        final String xml = xmlMapper.writeValueAsString(purchase);
```
pierre: Note that we won't test the serialization piece anymore.

bhelx:
I am now testing serialization and deserialization. This made me realize that netTerms should have been an integer so I also changed that from String to Integer.

bhelx:
@pierre this should be good to go. I should note this is technically a breaking change:
https://github.com/killbilling/recurly-java-library/pull/216/files#diff-bb0ba1f787cb09a5ae5821d7ea872814R115

PR #220, opened 2017-11-20
API Version 2.9

bhelx:
This PR contains the changes needed for API version 2.9.

1. Added the `authorize` endpoint for purchases. This endpoint allows collection with an external HPP. Currently only Adyen supported and we're still in beta.
2. Any subscription endpoint now accepts the `started_with_gift` parameter to filter out subscriptions that started with a gift card.

```java
final QueryParams params = new QueryParams();
params.put("started_with_gift", true);
```

There are breaking changes with the links in the XML API, however this library does not use the links so there should be no breaking changes this release.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -102,7 +102,7 @@ public class RecurlyClient {
     private static final Logger log = LoggerFactory.getLogger(RecurlyClient.class);
 
     public static final String RECURLY_DEBUG_KEY = "recurly.debug";
-    public static final String RECURLY_API_VERSION = "2.8";
+    public static final String RECURLY_API_VERSION = "2.9";
 
     private static final String X_RECORDS_HEADER_NAME = "X-Records";
     private static final String LINK_HEADER_NAME = "Link";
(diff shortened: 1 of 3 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -536,6 +536,22 @@ public Subscriptions getSubscriptions() {
...
+    /**
+     * Get all the subscriptions on the site given some sort and filter params.
+     * <p>
+     * Returns all the subscriptions on the site
+     *
+     * @param state {@link SubscriptionState}
+     * @param params {@link QueryParams}
+     * @return Subscriptions on the site
+     */
+    public Subscriptions getSubscriptions(final SubscriptionState state, final QueryParams params) {
```
bhelx: Seems like we were missing the signature of this method which accepts sorts and filters.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -536,6 +536,22 @@ public Subscriptions getSubscriptions() {
...
+        if (state != null) params.put("state", state.getType());
```
pierre: Nit: explicit braces.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -1498,6 +1514,25 @@ public Invoice previewPurchase(final Purchase purchase) {
...
+     + has been completed on an external source (e.g. Adyen's Hosted
+     + Payment Pages).
+     *
+     * <p>
+     * https://dev.recurly.com/docs/authorize-purchase
+     *
+     * @param purchase The purchase data
+     * @return The authorized invoice
+     */
+    public Invoice authorizePurchase(final Purchase purchase) {
```
pierre: Just my $0.02: the naming is a bit confusing as **authorize** is typically used in the concept of [auth/capture](https://en.wikipedia.org/wiki/Authorization_hold).

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: I'll drop those in

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: I agree. This is what Adyen calls the process so we went with that. The feature is still in beta so we may change it at some point in the future. Only a couple people can use this feature.

bhelx:
@pierre if you get a chance, could you do a release? Had a few people ask about getting this API version out in Java. Hope you have a nice thanksgiving!

pierre:
> @pierre if you get a chance, could you do a release? Had a few people ask about getting this API version out in Java.

0.13.0 released!

> Hope you have a nice thanksgiving!

You too, enjoy the break! 🎉

PR #230, opened 2018-02-07
Refactored AbstractTransactions to include failure_type

DBreitigan:
Refactored failureType out of PushTransaction and into AbstractTransaction since any type of transaction can have a failure type.


bhelx:
Thank you for taking the time to contribute this @DBreitigan !

Issue #214, opened 2017-08-28
Databinding problem with Jackson 2.9.0


geoffcallender:
The pom currently specifies jackson 2.5.0, but I have other REST interfaces that pushed me to upgrade jackson. I've found that recurly-java-library works fine with 2.8.10 but not with 2.9.0. 

The problem is that lazy loading fails. For example, after getting a Subscription, its Account href is null, its Invoice href is null, and so on. In a nutshell, 2.9.0 does not populate RecurlyObject's href.

Is there a known solution. A missing annotation or configuration option, perhaps?

pierre:
Jackson 2.9 isn't currently supported. See also the related discussion here: https://github.com/killbilling/recurly-java-library/issues/187#issuecomment-300287772.

A solution discussed in the past would be to provide shaded builds.

mmichalko:
I would also welcome shaded official version, since now I have to build this library on my own.

bhelx:
@mmichalko see #225. Would love to get some help testing it and getting it working.

mmichalko:
@bhelx added a comment there.

eodonohoe:
I believe putting @JsonProperty on the the RecurlyObject.setHref method fixes the issue with jackson-databind 2.9, need to keep the @JsonIgnore for on the getter.  From local testing this allows it to be read and not written.

eodonohoe:
FYI 
https://nvd.nist.gov/vuln/search/results?adv_search=true&cpe=cpe%3a%2fa%3afasterxml%3ajackson-databind%3a2.5.0

CVE vulnerabilities for jackson databind including 2.5.0:
CVE-2018-7489
CVE-2017-7525
CVE-2017-15095
CVE-2018-5968
CVE-2017-17485

bhelx:
Closed with #225

PR #238, opened 2018-03-30
Add JsonProperty to fix jackson 2.9 issue

eodonohoe:
Fix for killbilling/recurly-java-library#214 issue


bhelx:
Thanks @eodonohoe!

I assume this annotation is backwards compatible all the way to 2.5.X?

eodonohoe:
Yeah, I just added a simple check into TestAccount to verify it is present on deserialization and still not present in serialization for jackson 2.5.0.  I'm using jackson 2.9.4 locally and it works the same.

PR #246, opened 2018-04-18
Add support INR (indian rupees)

irphilli:
Added indian rupees to RecurlyUnitCurrency


bhelx:
Thanks for catching this @irphilli

PR #247, opened 2018-04-18
Add get/set currency to BillingInfo

irphilli:
Adds get/set for currency in billing info (allows users to specify the intended currency when setting billing info)


irphilli:
Updated!

PR #257, opened 2018-06-01
Add /transactions endpoint for invoices

bhelx:
This is for API version 2.13. This endpoint can be used to fetch all the transactions on the invoice. Normally they are embedded in the invoice, but in the case where there are more than 500, you must use this endpoint to paginate them.


bhelx:
Travis tests will pass when this goes to master.

PR #258, opened 2018-06-01
Allow programmer to set gateway code in purchase

bhelx:
For API version 2.13. Allows programmer to set the gateway for a purchase. If this changes, I will follow up with another PR on the `api_version_2_13` branch.


bhelx:
Travis tests will pass when this goes to master.

PR #263, opened 2018-06-26
API version 2.13

bhelx:
Changes for API version 2.13.


bhelx:
@pierre Can we do a release when you get the chance? There should not be any breaking changes

pierre:
Sure, will do today.

PR #266, opened 2018-07-03
Adds CustomField attribute on Account & Subscription

stepchud:


bhelx:
Closing and picking up with #267

PR #268, opened 2018-07-17
Test against multiple JDKs

bhelx:
Attempting to run the travis tests against multiple JDKs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,5 @@
 language: java
+jdk:
+  - oraclejdk7
+  - oraclejdk8
 script: ./travis-test.sh


pierre:
FWIW I originally had to make the code run on Java 6. Nowadays, if we don't have users on old JDK versions anymore, we could relax quite a few constraints in the code (and upgrade libraries, etc.).

Not sure if you have a way internally to poll users?

bhelx:
@pierre yes, we have the java version in the user agent. I don't have the stats yet, but we are going to look into which versions we will officially support and make sure we are running CI against them.

PR #270, opened 2018-08-01
fix autoRenew parameter

irphilli:
The setAutoRenew function in subscription has the wrong parameter name.


bhelx:
Thanks @irphilli 
Good catch

PR #272, opened 2018-08-13
Remove unnecessary query param "per_page" from non-list endpoints

### LINKED ISSUE #212: per_page=20 parameter on non-list endpoints
Calls are being made to non-list endpoints with a `per_page=20` attached to the end. We see examples including calls to:

`/v2/accounts/:account_code/balance?per_page=20`
`/v2/accounts/:account_code/redemption?per_page=20`
`/v2/accounts/:account_code/billing_info?per_page=20`

This is not causing any errors, the calls are still successful, but the parameter should be removed.

aaron-suarez:
Fixes #212

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Temp files and IDE files
 *.*#
+.vscode
 .idea
 *.iml
 *.ipr
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -203,7 +203,7 @@ public class RecurlyClient {
      * @return Accounts on success, null otherwise
      */
     public Accounts getAccounts() {
-        return doGET(Accounts.ACCOUNTS_RESOURCE, Accounts.class);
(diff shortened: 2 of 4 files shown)


aaron-suarez:
I was basing it off the docs, which typically list out the acceptable params (including `per_page`). Measured Units [does not have any acceptable params listed](https://dev.recurly.com/docs/list-measured-units) but that's probably an issue with the docs rather than the endpoint.

The transactions one isn't even listed on the docs page as a GET (the closest thing I could find was [this POST](https://dev.recurly.com/docs/enter-an-offline-payment-for-a-manual-invoice)). That being said, I can go ahead and add the QueryParams to those endpoints under the assumption that it's a docs issue.

aaron-suarez:
Added QueryParams to those endpoints [here](https://github.com/killbilling/recurly-java-library/pull/272/files#diff-8f900f0c920eccd0802341be3c6e3216R1685) and [here](https://github.com/killbilling/recurly-java-library/pull/272/files#diff-8f900f0c920eccd0802341be3c6e3216R968)

PR #280, opened 2018-09-14
Fix testCreateTrialExtensionCoupon integration test

bhelx:


bhelx:
If we do i imagine every tech company will need to hire one person to specifically find and fix all these bugs. (kind of like Y2K). Hopefully that person finds this commit easily :)

PR #293, opened 2018-10-24
Update refund calls to support credit memos

bhelx:
Deprecating current way of doing refunds for a single refund method
which now supports all possible options.

This is not a breaking change but does deprecate the current methods for doing refunds on Invoices.

In order to do refunds now, the programmer should use the `InvoiceRefund` object:

```java
final InvoiceRefund refundOptions = new InvoiceRefund();
refundOptions.setRefundMethod(RefundMethod.transaction_first);
refundOptions.setAmountInCents(100);
refundOptions.setCreditCustomerNotes("Credit Customer Notes");
refundOptions.setExternalRefund(true);
refundOptions.setPaymentMethod("credit_card");
final Invoice refundInvoice = recurlyClient.refundInvoice(invoice.getId(), refundOptions);
```

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -30,8 +30,6 @@ import com.ning.billing.recurly.model.BillingInfo;
 import com.ning.billing.recurly.model.Coupon;
 import com.ning.billing.recurly.model.Coupons;
 import com.ning.billing.recurly.model.CreditPayments;
-import com.ning.billing.recurly.model.CustomField;
-import com.ning.billing.recurly.model.CustomFields;
 import com.ning.billing.recurly.model.Errors;
 import com.ning.billing.recurly.model.GiftCard;
 import com.ning.billing.recurly.model.GiftCards;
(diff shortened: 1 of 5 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -30,8 +30,6 @@
 import com.ning.billing.recurly.model.Coupon;
 import com.ning.billing.recurly.model.Coupons;
 import com.ning.billing.recurly.model.CreditPayments;
-import com.ning.billing.recurly.model.CustomField;
-import com.ning.billing.recurly.model.CustomFields;
```
bhelx: These packages weren't used

src/main/java/com/ning/billing/recurly/model/Adjustment.java
```diff
@@ -106,6 +105,9 @@
     @XmlElement(name = "proration_rate")
     private BigDecimal prorationRate;
 
+    @XmlElement(name = "prorate")
```
bhelx: This actually needs to go on `AdjustmentRefund`

PR #300, opened 2018-11-20
Add hasOriginalInvoices() method to Invoice class

### LINKED ISSUE #295: Recurly issue found when preforming Credit refunds
Wanted to let you know about a recurly issue we found testing in our organization with NewCreditInvoiceNotification.
In the NewCreditInvoiceNotification, which is a refund, we need to get the original invoice using the invoice bundled with the notification. We lookup the invoice that came with the notification via the /invoices api and expect the original invoice (or just the original invoice id, i forget) to be returned with they payload. 
However, it is not returning this information for this type notification. We would expect it to have parity with a legacy return.
 
We have contacted Recurly about this issue. And they are looking into it. Just wanted to give you a heads-up in case you hear about it from other users.

aaron-junot:
Fixes #295 

Adds a method to check if there are any original invoices on an invoice received from the API or a webhook, so that an extra API call need not be made if there is no original invoice.

Example usage:
```java
if (invoice.hasOriginalInvoices()) {
    final Invoices originalInvoices = recurlyClient.getOriginalInvoices(invoice.getId());
}
```

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/Invoice.java
+++ b/src/main/java/com/ning/billing/recurly/model/Invoice.java
@@ -34,6 +34,9 @@ public class Invoice extends RecurlyObject {
     @XmlElement(name = "original_invoice")
     private Invoice originalInvoice;
 
+    @XmlElement(name = "original_invoices")
+    private Invoice originalInvoices;
+
     @XmlElement(name = "uuid")
     private String uuid;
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/ning/billing/recurly/TestRecurlyClient.java
```diff
@@ -1617,8 +1617,13 @@ public void testInvoices() throws Exception {
             Assert.assertEquals(refundInvoice.getTransactions().get(0).getAction(), "refund");
 
             // The refundInvoice should have an original_invoices of the original invoice
-            final Invoices originalInvoices = recurlyClient.getOriginalInvoices(refundInvoice.getId());
-            Assert.assertEquals(originalInvoices.get(0).getId(), invoice.getId());
+            if (refundInvoice.hasOriginalInvoices()) {
```
bhelx: Could we make this a unit test instead?

aaron-suarez: Done

Issue #244, opened 2018-04-12
Coupon's plans - Empty entries


jrg091:
I have a coupon assigned to all plans, in this case, when calling client's getCoupons, in the client's response I'm getting a list like: ["", "\n "] but in the API I get an empty array.

An example of the API response: 

```xml
<coupon href="https://remer.recurly.com/v2/coupons/code_01_test">
        <redemptions href="https://remer.recurly.com/v2/coupons/code_01_test/redemptions"/>
        <coupon_code>code_01_test</coupon_code>
        <name>coupon_01</name>
        <state>redeemable</state>
        ...
        <plan_codes type="array">
    </plan_codes>
        <a name="redeem" href="https://remer.recurly.com/v2/coupons/code_01_test/redeem" method="post"/>
    </coupon>
```

aaron-suarez:
I think I was able to reproduce. I did the following:
```java
            Coupons coupons = client.getCoupons();
            System.out.println(coupons.get(0).getPlanCodes());
```
And it printed this out:
```
[,
    ]
```
I've never seen an array-like object printed like that before, this will be interesting to fix.

aaron-suarez:
I'm going to let @bhelx handle this. I can't figure out what configuration of Jackson we need to be able to parse this correctly.

PR #302, opened 2018-11-21
coupon: fix plan_codes parsing

pierre:
This fixes https://github.com/killbilling/recurly-java-library/issues/244.

Any idea for the `must_apply_to_plans_or_non_plan_charges` error on Travis?

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/Coupon.java
+++ b/src/main/java/com/ning/billing/recurly/model/Coupon.java
@@ -488,6 +488,9 @@ public class Coupon extends RecurlyObject {
         if (couponCode != null ? !couponCode.equals(coupon.couponCode) : coupon.couponCode != null) {
             return false;
         }
+        if (planCodes != null ? !planCodes.equals(coupon.planCodes) : coupon.planCodes != null) {
+            return false;
+        }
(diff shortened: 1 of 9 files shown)


aaron-suarez:
This is an example of the XML being sent by `createCoupon`

```xml
<coupon xmlns=""><name>yzEBnzLFk3</name><coupon_code>yzebnzlfk3</coupon_code><applies_to_all_plans>false</applies_to_all_plans><discount_type>percent</discount_type><discount_percent>61</discount_percent><plan_codes><plan_code></plan_code></plan_codes></coupon>
```
Notice at the very end that it does not contain the plan_code in `<plan_codes><plan_code></plan_code></plan_codes>`

I'll look into it a little farther later if you don't get to it by then, but I suspect that `couponDataForPlan.setPlanCodes(planCodes);` is not successfully setting the plan code, or that there's something wrong with the serialization.

pierre:
@aaron-suarez Should be fixed! 😅

bhelx:
If it works, let's merge it! Fixing the general case problem seemed like a rabbit-hole to me and I think we should provide some relief on this problem for the time being.

bhelx:
Thanks @pierre !

PR #310, opened 2018-12-13
Update README based on current technologies

aaron-suarez:
Recurly has not supported TLS 1.1 since July 2018. See https://dev.recurly.com/page/tls-v10-deprecation

Recurly.js v4 does not require signatures using the private key. See https://dev.recurly.com/docs/recurlyjs


aaron-suarez:
Note: We should take out all of the signature signing code and such for the deprecated versions of RecurlyJS, but that can be its own PR

PR #321, opened 2019-02-12
Disable External Entities from XmlMapper

aaron-suarez:
Jackson version 2.7.4 and above disable XML external entity processing (XXE) by default. Since we're still using 2.5.0, we have to manually disable XXE.


aaron-suarez:
@pierre When you get a chance, would you mind cutting a release? We would like to get this change available for those who want it as soon as possible.

pierre:
Done!

aaron-suarez:
Thanks!

PR #327, opened 2019-03-11
Validate all API calls are being made to recurly.com

aaron-suarez:
The other Recurly clients all validate that the domain requests are being made to is "recurly.com". This adds a check to verify the same is true in this library.

Testing:

Initialize a RecurlyClient with the "subdomain" as `maliciousdomain.com/`. See how the client now blocks requests from being made to that malicious domain.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -90,6 +90,7 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.math.BigDecimal;
 import java.net.ConnectException;
+import java.net.URI;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.security.KeyManagementException;
@@ -101,6 +102,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.List;
+import java.util.Arrays;
 
 public class RecurlyClient {
 
@@ -119,6 +121,8 @@ public class RecurlyClient {
     private static final Pattern TAG_FROM_GIT_DESCRIBE_PATTERN = Pattern.compile("recurly-java-library-([0-9]*\\.[0-9]*\\.[0-9]*)(-[0-9]*)?");
 
     public static final String FETCH_RESOURCE = "/recurly_js/result";
+
+    private static final List<String> validHosts = Arrays.asList("recurly.com");
 
     /**
      * Checks a system property to see if debugging output is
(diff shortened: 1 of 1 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -120,6 +123,8 @@
 
     public static final String FETCH_RESOURCE = "/recurly_js/result";
 
+    private static final ArrayList<String> validHosts = newArrayList("recurly.com");
```
bhelx: Is there a reason we are using an ArrayList instead of an array?

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -2260,6 +2277,18 @@ protected AsyncHttpClient createHttpClient() throws KeyManagementException, NoSu
         return new AsyncHttpClient(builder.build());
     }
 
+    private void validateHost(String url) {
+        String host = URI.create(url).getHost();
+
+        // Remove the subdomain from the host
+        host = host.substring(host.indexOf(".")+1);
```
bhelx: Will this work for qa domains?

aaron-suarez: It will if you add `qaX.recurlyqa.com` to the validHosts array, rather than just `recurlyqa.com`

src/main/java/com/ning/billing/recurly/RecurlyClient.java
aaron-suarez: I've always used ArrayLists instead of arrays in Java because I find that the extra functionality is useful. In this case, I wanted to be able to use `[contains](https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#contains(java.lang.Object))` for the if check rather than loop through the array, but if there's a reason to use a plain old array instead of an ArrayList, I can loop through it in the check

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: What would that interface look like? It's a private variable.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
bhelx: Is the list intended to be modified? Also i think that relying on a dependency to create the new arraylist might be overkill.  Let's just use `Arrays.asList`, a literal representation, or something in the java standard library if we can. It's good practice to use the interface `List<String>` as a reference instead of the concrete class as well.

PR #329, opened 2019-03-12
Fix stackoverflow exception

### LINKED ISSUE #326: infinite loop in BillingInfo.toString()
`Account` prints `billingInfo` inside `toString()` method but the billing info is trying to print the `account` as well causing an infinite loop and `StackOverflowException`.

aaron-suarez:
Fixes #326 

Print just the account code if the account's billing info is set to the same object as the billing info to prevent an infinite when there's a circular reference.

Script:
```java
            final Account account = new Account();
            account.setAccountCode("b");

            final BillingInfo info = new BillingInfo();

            // create a circular reference
            account.setBillingInfo(info);
            info.setAccount(account);
            
            System.out.println(account); // now this does not throw an exception
```

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -747,9 +747,34 @@ public class RecurlyClient {
      * Please note: this API end-point may be used to import billing information without security codes (CVV).
      * Recurly recommends requiring CVV from your customers when collecting new or updated billing information.
      *
+     * @param accountCode recurly account id
      * @param billingInfo billing info object to create or update
      * @return the newly created or update billing info object on success, null otherwise
      */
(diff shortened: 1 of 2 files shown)


aaron-suarez:
This PR also deprecates the method which requires you to attach an Account object to a BillingInfo object. Here is an example of the new usage:
```java
client.open();

final BillingInfo billingInfo = new BillingInfo();

billingInfo.setFirstName("Aaron");
billingInfo.setLastName("Du Monde");
billingInfo.setAddress1("125 Paper Street");
billingInfo.setCity("Los Angeles");
billingInfo.setState("CA");
billingInfo.setZip("95312");
billingInfo.setCountry("US");
billingInfo.setPhone("555-555-5555");

client.createOrUpdateBillingInfo("x", billingInfo); // accountCode, billingInfo
```

PR #331, opened 2019-03-15
Transaction: subscription removed plus ability to pull subscriptions

demetrio812:
Hello, I've made the following changes:

- Transaction:
    - old subscription removed
    - new subscriptions property
- RecurlyClient: new getTransactionSubscriptions method

Let me know if it's ok

Thanks,
Dem

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/Transaction.java
+++ b/src/main/java/com/ning/billing/recurly/model/Transaction.java
@@ -33,8 +33,8 @@ public class Transaction extends AbstractTransaction {
     @XmlElement(name = "invoice")
     private Invoice invoice;
 
-    @XmlElement(name = "subscription")
-    private String subscription;
+    @XmlElement(name = "subscriptions")
+    private Subscriptions subscriptions;  // TODO remove
 
     @XmlElement(name = "uuid")
     private String uuid;
@@ -97,12 +97,10 @@ public class Transaction extends AbstractTransaction {
         this.invoice = invoice;
     }
 
-    public String getSubscription() {
-        return subscription;
-    }
+    public Subscriptions getSubscriptions() { return subscriptions; }
 
-    public void setSubscription(final Object subscription) {
-        this.subscription = stringOrNull(subscription);
+    public void setSubscription(final Subscriptions subscriptions) {
+        this.subscriptions = subscriptions;
     }
 
     public String getUuid() {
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/model/Transaction.java
```diff
@@ -33,8 +33,8 @@
     @XmlElement(name = "invoice")
     private Invoice invoice;
 
-    @XmlElement(name = "subscription")
-    private String subscription;
+    @XmlElement(name = "subscriptions")
+    private Subscriptions subscriptions;  // TODO remove
```
aaron-suarez: I played around with this for a little while. In the current implementation, if you receive a Transaction from the API that has no subscription attached, `transaction.getSubscriptions()` returns `null`, but if it does have some subscriptions, it returns an empty list `[]`.

I don't think this is particularly useful. If anything, a simple boolean like `hasSubscriptions()` would be all you need to check to know whether you should call `client.getTransactionSubscriptions()`. If you want to add that in, you can follow the pattern established in #300 

Also, there is no way to specify the subscription(s) when creating a transaction (see [the docs](https://dev.recurly.com/docs/create-transaction))

That being said, what I would like to see is the `subscriptions` property and all the related code removed completely, and just have the `getTransactionSubscriptions` method in the client because that's the really important piece to this.

You can add the method for `hasSubscriptions` if you like, but if you won't personally get use out of it, we can add it in when someone asks for it.

aaron-suarez:
Hey there @demetrio812! Do you think you're going to go ahead and make the requested changes or should we go ahead and close this out?

aaron-suarez:
@demetrio812 I'm going to go ahead and close this out. If you'd like to open another PR with the design choices mentioned above, someone will be able to review and merge. Thanks!

demetrio812:
Sorry @aaron-suarez haven't had much time lately, will try to close this asap, will send another one, thanks

src/main/java/com/ning/billing/recurly/model/Transaction.java
joannasese: Hi @aaron-suarez. I'm picking this up. See https://github.com/killbilling/recurly-java-library/pull/374.

joannasese: Hi @aaron-suarez. I'm picking this up. See https://github.com/killbilling/recurly-java-library/pull/374

PR #332, opened 2019-03-18
Throw errors when input is empty

aaron-suarez:
If you enter an empty string into certain RecurlyClient methods, the API will return all of that resource. E.g.,
```java
Account account = client.getAccount("");
```
This can lead to information disclosure in certain situations and should not be allowed.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -307,6 +307,9 @@ public class RecurlyClient {
      * @return account object on success, null otherwise
      */
     public Account getAccount(final String accountCode) {
+        if (accountCode == null || accountCode.isEmpty())
+            throw new RuntimeException("accountCode cannot be empty!");
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -307,6 +307,9 @@ public Integer getCouponsCount(final QueryParams params) {
      * @return account object on success, null otherwise
      */
     public Account getAccount(final String accountCode) {
+        if (accountCode == null || accountCode.isEmpty())
+            throw new RuntimeException("accountCode cannot be empty!");
```
aaron-suarez: I chose to throw `RuntimeException`s so we don't have to change the method signature. This way, it's not a breaking change

aaron-suarez: Also note that the null check is only necessary because this would throw a NullPointerException if the input was null, not because the API would return any information (unless you happened to have an account with account code "null". But in that situation, you should pass in that string rather than an actual `null`).

PR #350, opened 2019-07-16
Fix multiple coupon codes

### LINKED ISSUE #349: Malformed XML mapped when adding multiple coupon codes
Hello,

**Describe the bug**

The issue I'm facing is that: whenever I try to call the "create subscription" or "preview subsription" with a subscription that we have set a list of coupons into always return the XML malformed error. It complains that tag <coupon_codes> should contain only sub tag <coupon_code> (code without 's').
Recurly tech guy provided me the malformed part:

```
<coupon_codes>   
     <coupon_codes>dungdiscount</coupon_codes>    
     <coupon_codes>ben_code_10</coupon_codes>
</coupon_codes>
```

which is supposed to be below instead:
```
<coupon_codes>   
     <coupon_code>dungdiscount</coupon_code>    
     <coupon_code>ben_code_10</coupon_code>
</coupon_codes>
```

**To Reproduce**
 
```      // Apply multiple coupon at the time of subscription creation
            List<String> couponCodes = new ArrayList<String>();
            couponCodes.add("dungdiscount");
            couponCodes.add("ben_code_10");
            subscriptionData.setCouponCodes(couponCodes);
            final Subscription subscriptionPreview = recurlyClient.previewSubscription(subscriptionData);
```

https://v2uploads.zopim.io/v/x/3/vx3xUhLgBy3QROz61Nhq6O0TdxoYyN04/385fdfd40a31e30d8119e778b74d8e387a75b586.png

** Possible root cause **
Found out something that  might be the root cause, its this class in the java client library: 

`com.ning.billing.recurly.model.Subscription`

    
``` 
    @XmlList
    @XmlElementWrapper(name = "coupon_codes”) —> This line of code made the tag <coupon_codes> reappear as subtags of that same tag.
    private List<String> couponCodes;
```

Changing it below will make the XML to be formatted as expected:
   
 ``` 
    @XmlElementWrapper(name = "coupon_codes")
    @XmlElement(name="coupon_code”) —> this line of code will allow the tag <coupon_code> appears as sub tag of <coupon_codes>
    private List<String> couponCodes;
```

**Expected behavior**
XML is formed correctly and the API returned a subscription object with corresponding data.
<coupon_codes>   
     <coupon_code>dungdiscount</coupon_code>    
     <coupon_code>ben_code_10</coupon_code>
</coupon_codes>

**Your Environment**

- Which version of this library are you using?
recurly-java-library: 0.28.0
RECURLY_API_VERSION=2.21

### PR DESCRIPTION
**Pull request by aaron-suarez**
Fixes #349 

The root cause of the problem described in #349 is as follows:

The Recurly v2 API Create Subscription and Preview Subscription endpoints accept both a single coupon code, e.g.
```xml
<coupon_code>my-coupon</coupon_code>
```
and multiple coupon codes, e.g.
```xml
<coupon_codes>
  <coupon_code>my-coupon</coupon_code>
  <coupon_code>my-other-coupon</coupon_code>
<coupon_codes>
```

As such, there are two fields in the `Subscription` class:
```java
private String couponCode;
```
and
```java
private List<String> couponCodes;
```

FasterXML Jackson only allows for one field to be associated with the tag `<coupon_code>` as documented by this issue https://github.com/FasterXML/jackson-dataformat-xml/issues/192

This PR gets around this limitation by having Jackson ignore the field `couponCode` and when the programmer uses `setCouponCode()`, adding that as the first element of the list for `couponCodes`. It should be noted that both of the following are accepted by the Recurly API Subscription endpoints as equivalent:
```xml
<coupon_code>my-coupon</coupon_code>
```
```xml
<coupon_codes>
  <coupon_code>my-coupon</coupon_code>
<coupon_codes>
```

It's kind of gross to have to do it this way, but I could not find another suitable workaround with this limitation of Jackson.

### DIFF
--- a/src/test/java/com/ning/billing/recurly/model/TestSubscription.java
+++ b/src/test/java/com/ning/billing/recurly/model/TestSubscription.java
@@ -19,6 +19,7 @@ package com.ning.billing.recurly.model;
 
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.util.ArrayList;
 
 import com.ning.billing.recurly.TestUtils;
 import org.joda.time.DateTime;
@@ -27,6 +28,7 @@ import org.testng.annotations.Test;
 
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotEquals;
+import static org.testng.Assert.assertTrue;
 
 public class TestSubscription extends TestModelBase {
 
@@ -92,6 +94,11 @@ public class TestSubscription extends TestModelBase {
         verifyPaginationData(subscription);
         verifyPendingSubscription(subscription);
         Assert.assertEquals(subscription.getAddOns().size(), 0);
+
+        ArrayList<String> codes = new ArrayList<String>();
+        codes.add("123");
+        codes.add("abc");
+        assertEquals(subscription.getCouponCodes(), codes);
     }
 
     @Test(groups = "fast")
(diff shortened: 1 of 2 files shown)


src/test/java/com/ning/billing/recurly/model/TestSubscription.java
```diff
@@ -92,6 +94,11 @@ public void testDeserialization() throws Exception {
         verifyPaginationData(subscription);
         verifyPendingSubscription(subscription);
         Assert.assertEquals(subscription.getAddOns().size(), 0);
+
+        ArrayList<String> codes = new ArrayList<String>();
+        codes.add("123");
+        codes.add("abc");
+        assertEquals(subscription.getCouponCodes(), codes);
```
aaron-suarez: It should be noted that this check is ultra-paranoid, because the v2 API never returns `coupon_codes` in the Subscription object. But since the XML was like this, I thought it would be nice to check that it still works.

PR #351, opened 2019-07-25
Fix duplicate apostrophe in Account's toString

walfie:
The `toString` implementation had an extra apostrophe for parent account
code, so it would show up as: `parent_account_code=''null'`


aaron-suarez:
Thanks for this @walfie!

PR #352, opened 2019-07-30
Add method for reopening a closed account

walfie:
Also adds the `closed_at` field from the XML to the Account object,
and fixes up some inconsistencies (underscores instead of camelcase)
in the `toString` implementation.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -353,6 +353,19 @@ public class RecurlyClient {
     }
 
     /**
+     * Reopen Account
+     * <p>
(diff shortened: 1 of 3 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -352,6 +352,19 @@ public void closeAccount(final String accountCode) {
...
     }
 
+    /**
+     * Reopen Account
+     * <p>
+     * Transitions a closed account back to active.
+     *
+     * @param accountCode recurly account id
+     */
+    public Account reopenAccount(final String accountCode) {
```
bhelx: 👍 Good catch. thank you!

PR #356, opened 2019-08-15
Changes for Vertex Customers

bhelx:
This change only applies to customers with a Vertex integration (taxes).

- `Adjustment#surchargeInCents`
- `Invoice#surchargeInCents`
- `Invoice#taxClassification`


.travis.yml
```diff
@@ -1,2 +1,4 @@
 language: java
 script: ./scripts/test travis
+jdk:
```
bhelx: Apparently travis changed on us so now I need to force this to JDK 1.8

PR #358, opened 2019-08-21
MOTO transactions

bhelx:
As a part of PSD2, there is an exemption to exclude MOTO transactions from SCA scope.
In order to pass this indicator to the gateways, merchants need a mechanism via API
to inform Recurly of these transactions. These transactions will be initiated from
within a merchants customer service / support admin portal UI, where the customer
is not authenticated to be able to complete an SCA flow.

This is exposed to the programmer as transaction_type on many of the resources. Set the value to moto to mark the transaction as MOTO. Ex use with purchases:

```java
final Purchase purchaseData = new Purchase();
purchaseData.setCurrency("USD");
purchaseData.setTransactionType("moto");
purchaseData.setAccount(accountData);
purchaseData.setSubscriptions(subscriptionsData);
InvoiceCollection collection = recurlyClient.purchase(purchaseData);
```

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -1327,6 +1327,18 @@ public class RecurlyClient {
     }
 
     /**
+     * Force collect an invoice
+     *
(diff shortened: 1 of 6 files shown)


pierre:
0.30.0 released :-)

PR #364, opened 2019-10-17
Add shipping address to purchase

joannasese:
- Add shipping address to Purchase class
- Add shipping address tests to TestPurchase 
- Add shipping address to Purchase builder


bhelx:
Let's add a shipping address to the purchase builder for the tests: https://github.com/killbilling/recurly-java-library/blob/master/src/test/java/com/ning/billing/recurly/TestUtils.java#L951

Something like:

```java
purchase.setShippingAddress(createRandomShippingAddress(seed));
```

This will also ensure that your shippingaddress is being properly considered in the hashcode calculation.

PR #381, opened 2020-02-21
Fix getCouponRedemptionsByInvoice

slisaasquatch:
The `getCouponRedemptionsByInvoice` method for some reason uses `/subscription` instead of `/subscriptions`.


joannasese:
@ slisaasquatch thanks for reporting. We'll take a look.

joannasese:
👍

PR #382, opened 2020-03-06
Add URL encoding to URL path segments

slisaasquatch:
I've run into an issue using this library where account codes with spaces and special characters wouldn't work. And the apparent cause was that URL path segments are not getting URL encoded. So I went ahead and added URL encoding to all the path segments. The behavior should not change for Strings without special characters.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -2669,7 +2669,7 @@ public class RecurlyClient {
     private static String urlEncode(String s) {
         try {
             return URLEncoder.encode(s, Charsets.UTF_8.name())
-                    .replace("+", "%20").replace("*", "%2A").replace("%7E", "~");
+                    .replace("+", "%20").replace("*", "%2A");
         } catch (UnsupportedEncodingException e) {
             throw new RuntimeException(e); // should not happen
         }
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -2660,12 +2660,16 @@ public class RecurlyClient {
         return matcher.find() ? matcher.group(1) : null;
     }
 
+    /**
+     * RFC 3986 URL encoding
+     */
     private static String urlEncode(String s) {
-    	try {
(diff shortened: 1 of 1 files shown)


joannasese:
@slisaasquatch sorry I missed this! I'll take a look and will contact you if I have any questions.

joannasese:
@slisaasquatch thanks as always for your contributions. We haven't merged this quite yet because we're interested in approaching this issue through utilizing a standard url encoder and avoiding string replacement if possible. Still on the radar, though! If you have further input, please chime in.

slisaasquatch:
So how do you like the idea of adding apache commons codec as a dependency? I used string replace because it avoids adding another dependency.

joannasese:
You're right to want to avoid adding another dependency. I'll talk this over with the team.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -2660,4 +2662,17 @@ String getVersionFromGitRepositoryState(final Properties gitRepositoryState) {
...
     }
 
+    /**
+     * RFC 3986 URL encoding. The vanilla {@link URLEncoder} does not work since
+     * Recurly does not decode '+' back to ' '.
+     */
+    private static String urlEncode(String s) {
+        try {
+            return URLEncoder.encode(s, Charsets.UTF_8.name())
+                    .replace("+", "%20").replace("*", "%2A").replace("%7E", "~");
```
joannasese, line 2672: How about we switch this to `.replace("+", "%20")` as it appears the only problematic character is the space.

slisaasquatch: So just deleted the following `.replace("*", "%2A").replace("%7E", "~")`?

joannasese: Actually, only remove `.replace("%7E", "~")`.  Motivation being we want to keep the string replacement to a minimum, and since we're not percent-encoding tildes to begin with, I'm not seeing a need for `.replace("%7E", "~")`. As for `.replace("*", "%2A")`, you are correct - that should be included.

slisaasquatch: OK just updated it.

PR #383, opened 2020-03-11
Enable item-backed add-ons

joannasese:
To create item-backed add-on:
```java
final AddOn addon1 = new AddOn();
addon1.setItemCode(item.getItemCode());
final RecurlyUnitCurrency price = new RecurlyUnitCurrency();
price.setUnitAmountUSD(10);
addon1.setUnitAmountInCents(price);

final AddOn addon2 = client.createPlanAddOn(plan.getPlanCode(), addon1);
```

To get item-backed add-on:
```java
final AddOn addon = client.getAddOn(plan.getPlanCode(), item.getItemCode());
```

To update item-backed add-on:
```java
AddOn addOnData = new AddOn();
addOnData.setDefaultQuantity(5);

final AddOn updatedAddOn = client.updateAddOn(plan.getPlanCode(), addon.getAddOnCode(), addOnData);
```

When add-on has an associated item, only the following attributes are updatable:
```java
unit_amount_in_cents
default_quantity
display_quantity_on_hosted_page
optional
```

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/AddOn.java
+++ b/src/main/java/com/ning/billing/recurly/model/AddOn.java
@@ -33,6 +33,9 @@ public class AddOn extends AbstractAddOn {
     @XmlElement(name = "name")
     private String name;
 
+    @XmlElement(name = "item_code")
+    private String itemCode;
+
     @XmlElement(name = "display_quantity_on_hosted_page")
     private Boolean displayQuantityOnHostedPage;
 
@@ -71,6 +74,14 @@ public class AddOn extends AbstractAddOn {
         this.name = stringOrNull(name);
     }
 
+    public String getItemCode() {
+        return itemCode;
+    }
+
+    public void setItemCode(final Object itemCode) {
+        this.itemCode = stringOrNull(itemCode);
+    }
+
     public Boolean getDisplayQuantityOnHostedPage() {
         return displayQuantityOnHostedPage;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/ning/billing/recurly/model/AddOn.java
```diff
@@ -92,7 +103,9 @@ public RecurlyUnitCurrency getUnitAmountInCents() {
     }
 
     public void setUnitAmountInCents(final RecurlyUnitCurrency unitAmountInCents) {
```
bhelx: I believe we want this to be `Object` not `RecurlyUnitCurrency`. It's the `build` method's job to cast it.

bhelx: I believe we want this to be Object not RecurlyUnitCurrency. It's the build method's job to cast it.

src/main/java/com/ning/billing/recurly/model/AddOn.java
```diff
@@ -92,7 +103,9 @@ public RecurlyUnitCurrency getUnitAmountInCents() {
...
-        this.unitAmountInCents = unitAmountInCents;
+        // this.unitAmountInCents = unitAmountInCents;
```
bhelx: Let's remove this comment if we don't need it

joannasese: Done! Was sloppy that I left it there.

PR #393, opened 2020-06-19
Add BACS support

joannasese:
Added `sort_code` and `type` fields to support BACS payment type.

Previously, the `type` field was a xml attribute with a protected setter. When BACS is the payment option, however, `type` must accept `"bacs"` as a value.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/BillingInfo.java
+++ b/src/main/java/com/ning/billing/recurly/model/BillingInfo.java
@@ -31,8 +31,11 @@ public class BillingInfo extends RecurlyObject {
     @XmlTransient
     public static final String BILLING_INFO_RESOURCE = "/billing_info";
 
-    @XmlAttribute(name = "type")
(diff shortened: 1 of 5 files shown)


src/main/java/com/ning/billing/recurly/model/BillingInfo.java
```diff
@@ -31,7 +31,7 @@
     @XmlTransient
     public static final String BILLING_INFO_RESOURCE = "/billing_info";
 
-    @XmlAttribute(name = "type")
+    @XmlElement(name = "type")
     private String type;
```
bhelx, line 35: We need to preserve the xml-attribute `type`

```
    @XmlElement(name = "type")
    private String type;

    @XmlAttribute(name = "type")
    private String attributeType;

    public void setType(final Object type) {
        this.type = stringOrNull(type);
    }

    public String getType() {
        return this.type == null ? this.attributeType : this.type;
    }
```

PR #394, opened 2020-06-19
Set content-length header to 0 when entity is null

thanhedman:
Requests for `null` entity operations (like `resumeSubscription`) are now rejected by the Recurly v2 API with status code `411`. This is because content-length header is not set in the case where there is no entity.

This PR sets the content-length header to 0 when the entity is null.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -81,6 +81,7 @@ import com.ning.billing.recurly.util.http.SslUtils;
 
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
+import org.apache.http.HttpEntityEnclosingRequest;
 import org.apache.http.NoHttpResponseException;
 import org.apache.http.ParseException;
 import org.apache.http.client.config.RequestConfig;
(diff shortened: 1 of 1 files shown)


thanhedman:
Sample error:
```Got Recurly API Exception RecurlyAPIError{description='null', symbol='null', details='null', httpStatusCode='411', responseMetadata='ResponseMetadata{requestId=null, cfRay=5a5aa65fa92ce0d6-IAD, statusCode=411}'}, Description: null```

joannasese:
@thanhedman Thanks for reporting. We're looking into this and will respond soon.

joannasese:
@thanhedman which version of the library are you using?

thanhedman:
@joannasese Discovered in production with `0.27.0`, reproduced with `0.34.0`.

thanhedman:
If you're a Recurly employee, I believe there is also back-channel communication going on about this, via Jeff Sheldon.

joannasese:
Hi @thanhedman. After looking into this, we found that the issue with content-length headers is addressed through an [Apache upgrade](https://github.com/killbilling/recurly-java-library/pull/390) that has yet to be released. We'll get `0.35.0` out shortly and will close out this PR after the release has been cut.

slisaasquatch:
Hi @thanhedman, I recently switched out the underlying HTTP client of this library from an outdated ning AHC to Apache HC4. I'm pretty sure that Apache HC4 will fix this issue. So basically the version you are having issues with is still using the old HTTP client. So I don't think this pull request is necessary.

thanhedman:
@joannasese @slisaasquatch that all makes sense! I saw the lib changes on `master:HEAD` vs `0.34.0` and thought they might fix the issue, but also thought that maybe that larger revision had more obstacles to release. I'll close this now!

bhelx:
@slisaasquatch thanks again! that PR is paying dividends.

PR #395, opened 2020-06-23
URL encoding cleanup

slisaasquatch:
I recently noticed that Apache Commons Codec is a transitive dependency of Apache HttpClient, so I switched the not-so-great String replace implementation of URL encoding to using the `URLCodec` class in Apache Commons Codec, which is a lot faster than Java's built-in `URLEncoder` even without the String replace.

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -2544,7 +2544,7 @@ public class RecurlyClient {
                 return null;
             }
 
-            final Header locationHeader = response.getFirstHeader("Location");
+            final Header locationHeader = response.getFirstHeader(HttpHeaders.LOCATION);
             String location = locationHeader == null ? null : locationHeader.getValue();
             if (clazz == Coupons.class && location != null && !location.isEmpty()) {
                 final RecurlyObjects recurlyObjects = new Coupons();
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -2562,7 +2562,7 @@ public class RecurlyClient {
             }
 
             final Header locationHeader = response.getFirstHeader(HttpHeaders.LOCATION);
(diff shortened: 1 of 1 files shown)


joannasese:
Hi @slisaasquatch. Thank you once again for your contribution! We'll get this reviewed and if everything checks out, we'll make a push to include it in the upcoming release scheduled for the end of this month.

src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -2691,16 +2708,12 @@ String getVersionFromGitRepositoryState(final Properties gitRepositoryState) {
...
      */
     private static String urlEncode(String s) {
-        try {
-            return URLEncoder.encode(s, Charsets.UTF_8.name())
-                    .replace("+", "%20").replace("*", "%2A");
-        } catch (UnsupportedEncodingException e) {
-            throw new RuntimeException(e); // should not happen
-        }
+        return new String(URLCodec.encodeUrl(RFC_3986_SAFE_CHARS, s.getBytes(Charsets.UTF_8)),
+                Charsets.UTF_8);
```
bhelx, line 2716: We really appreciate you cleaning up this code. Is there not a method in the Apache library or the java standard library that implements this? Did you find some parts that were missing from the previous implementation? Or is this just about making sure we follow the full spec?

slisaasquatch: This is the only library that does RFC 3986 standard URL encoding that I know of. The previous implementation works but it uses String replace, which uses regex (which we don't need), and that means it is slow. This implementation is actually faster than Java's built-in `URLEncoder` and it's already a transitive dependency.

slisaasquatch: I agree that ideally manual URL encoding like this should be avoided. We should be using the `URIBuilder` library Apache HC provides, but that will require major refactoring of the code and I don't know if it's worth it.

bhelx: No problem, thanks for the explanation!

PR #402, opened 2020-07-02
Cache user agent + reduce Guava beta API usage

slisaasquatch:
This pull request has 2 things.
1. The user agent is now being cached. Before whenever you created a new instance of `RecurlyClient`, it would unnecessarily perform some file IO to build the user agent, and you would get the same result every time.
2. I've replaced the usage of the `com.google.common.io.Resources` class, which is still marked as `@Beta` in Guava [as of July 1, 2020](https://github.com/google/guava/blob/d44669df2f1540ae52c278181377f0adea88f8fb/guava/src/com/google/common/io/Resources.java#L45). I know this library uses some other `@Beta` Guava class like `BaseEncoding` and `StandardSystemProperty`, but those classes are now out of beta in the latest version of Guava, so they should be stable.


joannasese:
Closing, as this has been merged through #402

Issue #405, opened 2020-07-15
Null Pointer Exception in toString() of BillingInfo


Shivani-SinghDeo:
**BillingInfo** class method **toString()** is giving Null Pointer Exception as the account is not null but the **account.getBillingInfo()** is coming as null while updating the billing info of an account

BillingInfo updatedBillingInfo = recurlyClient.createOrUpdateBillingInfo(accountCode, billingInfo);

Pass the account code and the billing info with the token. When printing updatedBillingInfo it would give NPE.

Code giving NPE : 

`if (account != null && account.getBillingInfo().equals(this)) {
            sb.append("{account='").append(account.getAccountCode()).append('\'');
        }`

**Environment**

Library Version - 0.27.0
Java Version - Java 8

Shivani-SinghDeo:
@joannasese Adding this code worked 

`Account account = getAccount();
		if (account != null && account.getBillingInfo() != null && this.getHref() != null
				&& this.getHref().equals(account.getBillingInfo().getHref())) {
			sb.append("{account='").append(account.getAccountCode()).append('\'');
		} else {
			sb.append("{account='").append(account).append('\'');
		}`

Shivani-SinghDeo:
@joannasese Can you please review this change - https://github.com/killbilling/recurly-java-library/pull/406

joannasese:
Hi @Shivani-SinghDeo. Thanks for the contribution! I was able to reproduce the error and am reviewing your PR.

PR #406, opened 2020-07-15
Fix Null Pointer Exception in toString() of BillingInfo #405

Shivani-SinghDeo:


slisaasquatch:
I just want to point out that the indentation is inconsistent with the rest of the project 🙈

src/main/java/com/ning/billing/recurly/model/BillingInfo.java
```diff
@@ -496,12 +496,14 @@ public String toString() {
...
         // See https://github.com/killbilling/recurly-java-library/issues/326
-        if (account != null && account.getBillingInfo().equals(this)) {
-            sb.append("{account='").append(account.getAccountCode()).append('\'');
-        }
-        else {
-            sb.append("{account='").append(account).append('\'');
-        }
+	// See https://github.com/killbilling/recurly-java-library/issues/405
+        Account account = getAccount();
+		if (account != null && account.getBillingInfo() != null && this.getHref() != null
```
joannasese, line 501: Hi @Shivani-SinghDeo. Apologies for the delay here. Because your PR touches https://github.com/killbilling/recurly-java-library/issues/326, I needed to check for unintended side affects from replacing `account.getBillingInfo().equals(this)` with `account.getBillingInfo() != null`. 

Can you explain your inclusion of `this.getHref() != null && this.getHref().equals(account.getBillingInfo().getHref())`? I'm not seeing the need for it.

src/main/java/com/ning/billing/recurly/model/BillingInfo.java
```diff
@@ -496,12 +496,14 @@ public String toString() {
 
         // Prevent infinite loop when printing account.
         // See https://github.com/killbilling/recurly-java-library/issues/326
```
joannasese, line 498: Please add memo:
```
// Prevent Null Pointer Exception when printing updated billing info
// See https://github.com/killbilling/recurly-java-library/issues/405
```

src/main/java/com/ning/billing/recurly/model/BillingInfo.java
Shivani-SinghDeo: @joannasese `account.getBillingInfo()` only returns the href in the BillingInfo object, rest all fields are null in the object  so the equals check would never become true. So it would be safe to check the href instead of the entire object.

Issue #185, opened 2017-04-30
doGET does not distinguish between "not found" and timeout.


geoffcallender:
If the RecurlyClient suffers timeout, it logs "Execution error" and returns null. To the caller, it's indistinguishable from "not found". For example, when I tried to find an Account...  

```
	Account account = getRecurlyClient().getAccount(accountCode);
```

...and it timed out, a null account was returned rather than an exception. The following was logged (using version 0.8.0)...  

```
ERROR [Timer-7] (RecurlyClient.java:1483) - Execution error
java.util.concurrent.ExecutionException: java.net.ConnectException: connection timed out: goxpro-dev.recurly.com/104.16.113.53:443
	at com.ning.http.client.providers.netty.future.NettyResponseFuture.abort(NettyResponseFuture.java:232)
	at com.ning.http.client.providers.netty.request.NettyConnectListener.onFutureFailure(NettyConnectListener.java:151)
	at com.ning.http.client.providers.netty.request.NettyConnectListener.operationComplete(NettyConnectListener.java:159)
	at org.jboss.netty.channel.DefaultChannelFuture.notifyListener(DefaultChannelFuture.java:409)
	at org.jboss.netty.channel.DefaultChannelFuture.notifyListeners(DefaultChannelFuture.java:400)
	at org.jboss.netty.channel.DefaultChannelFuture.setFailure(DefaultChannelFuture.java:362)
	at org.jboss.netty.channel.socket.nio.NioClientBoss.processConnectTimeout(NioClientBoss.java:142)
	at org.jboss.netty.channel.socket.nio.NioClientBoss.process(NioClientBoss.java:83)
	at org.jboss.netty.channel.socket.nio.AbstractNioSelector.run(AbstractNioSelector.java:337)
	at org.jboss.netty.channel.socket.nio.NioClientBoss.run(NioClientBoss.java:42)
	at org.jboss.netty.util.ThreadRenamingRunnable.run(ThreadRenamingRunnable.java:108)
	at org.jboss.netty.util.internal.DeadLockProofWorker$1.run(DeadLockProofWorker.java:42)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
[...]

PR #411, opened 2020-07-28
Re-throw SSLException as ConnectionErrorException rather than return null

thanhedman:
## What

- Add `SSLException` to the subset of `IOException` which throw `ConnectionErrorException` rather than warn and return null

## Why

- Currently, there is no way to handle `SSLException`, which may be ephemeral and retried, because the exception is swallowed
- Conceptually, `SSLException` is an ephemeral error at the connection level, so the inclusion makes sense
- Practically, most `IOException` instances which aren't re-thrown in this branch are XML-related

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -105,6 +105,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
+import javax.net.ssl.SSLException;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -2491,7 +2492,7 @@ public class RecurlyClient {
             return callRecurlyXmlContent(builder, clazz);
         } catch (IOException e) {
             if (e instanceof ConnectException || e instanceof NoHttpResponseException
-                    || e instanceof ConnectTimeoutException) {
+                    || e instanceof ConnectTimeoutException || e instanceof SSLException) {
                 // See https://github.com/killbilling/recurly-java-library/issues/185
                 throw new ConnectionErrorException(e);
             }


slisaasquatch:
Just a comment. I'm not even sure if it ever makes sense for that method to return null. I kept it when I did the migration because I wanted to keep the old behavior.

pierre:
FYI https://github.com/killbilling/recurly-java-library/issues/185#issuecomment-300291294 might be useful to provide context on the original design.

thanhedman:
@slisaasquatch I was thinking that I would go for the minimally invasive change in case anyone is depending on the null for their implementation, but I'm definitely open to re-throwing everything (and keeping the conditional for those that should be re-thrown as `ConnectionErrorException`) if that's the direction y'all think is best

thanhedman:
@pierre I'll read up on that and think about how @slisaasquatch 's suggestion would intersect

pierre:
> @pierre I'll read up on that and think about how @slisaasquatch 's suggestion would intersect

Sounds good!

In case it helps, in the [Kill Bill Java client](https://github.com/killbill/killbill-client-java#usage), we wrap all exceptions and only return `null`/empty list in case of 204 and 404: https://github.com/killbill/killbill-client-java/blob/79d1159661770342aed5d712a49935883e29888d/src/main/java/org/killbill/billing/client/KillBillHttpClient.java#L323-L342

cdadia:
@pierre Any chance we could either get this PR reviewed or change the behavior based on your comment here
https://github.com/killbilling/recurly-java-library/issues/185#issuecomment-300291294

SSLException is a real issue and Recurly teams suggestion is to basically just catch and handle but the current implementation does not allow for that.

chrissrogers:
In my opinion, this is scoped reasonably within the `ConnectionErrorException` and is a good change. Agreed with the direction from @thanhedman to avoid changing behavior in case there are cases where folks rely on the null return.. and if it's desirable to change that paradigm, then it can be done in a subsequent PR with a wider scope.

PR #422, opened 2021-01-07
Change test cc year so it's no longer expired

joannasese:
Integration tests were failing for all tests that involved billing info with cc, as expiry date was configured as 11/20. I changed the expiry year to 2049 to repair the tests.


douglasmiller (approved):
👍

Issue #419, opened 2020-11-27
[Feature Request] Support Authorization Cancellation


zocomd:
Hi, 

Can we please add support for Authorization Cancellation? It should map to [v2/purchases/cancel API](https://docs.recurly.com/docs/auth-and-capture#section-auth-and-capture). I see we already support [authorizePurchase()](https://t.ly/LQlU). If we already have a way to cancel an authorization, please let me know too. Thanks in advance.

_Warning_: Github issues are not an official support channel for Recurly. If you have an urgent request we suggest you contact us through an official channel: support@recurly.com or https://recurly.zendesk.com

joannasese:
Hi @zocomd. Apologies for missing this! We'll see what we can do about adding this endpoint. I'll keep this issue open for now.

joannasese:
Hi @zocomd. We've added support for capture and cancel authorizations. These updates will be included in the next release.

PR #424, opened 2021-01-21
Add cancel and capture purchase functionality

joannasese:
- Also updated links to documentation
- Added tests for capture and cancel purchase

Addresses https://github.com/killbilling/recurly-java-library/issues/419

### DIFF
--- a/src/main/java/com/ning/billing/recurly/RecurlyClient.java
+++ b/src/main/java/com/ning/billing/recurly/RecurlyClient.java
@@ -2174,7 +2174,7 @@ public class RecurlyClient {
     /**
      * Purchases endpoint
      * <p>
-     * https://dev.recurly.com/docs/create-purchase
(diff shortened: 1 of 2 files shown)


src/main/java/com/ning/billing/recurly/RecurlyClient.java
```diff
@@ -2428,7 +2456,11 @@ private InputStream callRecurlySafeGetPdf(String url) {
     private <T> T doPOST(final String resource, final RecurlyObject payload, final Class<T> clazz) {
         final String xmlPayload;
         try {
-            xmlPayload = RecurlyObject.sharedXmlMapper().writeValueAsString(payload);
+            if (payload != null) {
+                xmlPayload = RecurlyObject.sharedXmlMapper().writeValueAsString(payload);
+            } else {
```
joannasese, line 2461: Handles case where payload is `null`, as is the case with `CapturePurchase` and `CancelPurchase`

PR #433, opened 2021-10-08
Client library update for V2 Dunning Campaign API enhancements

amandamfielding:
* Add `dunning_campaign_id` to `Account`, `Invoice`, and `Plan` 
* PUT v2/dunning_campaigns/{dunning_campaign_id}/bulk_update

### DIFF
--- a/src/main/java/com/ning/billing/recurly/model/DunningCycle.java
+++ b/src/main/java/com/ning/billing/recurly/model/DunningCycle.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2010-2014 Ning, Inc.
+ * Copyright 2014-2015 The Billing Project, LLC
+ *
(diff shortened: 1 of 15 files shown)


src/main/java/com/ning/billing/recurly/model/DunningCycle.java
```diff
@@ -0,0 +1,160 @@
...
+    private Boolean failInvoice;
+
+    @XmlElement(name = "total_dunning_days")
+    private Integer totalDunningDays;
+
+    @XmlElement(name = "total_recycling_days")
+    private Integer totalRecyclingDays;
+
+    @XmlElement(name = "version")
+    private String version;
```
sunnyw1212, line 57: should version be an Integer?

amandamfielding: ah you're right, will update

PR #483, opened 2024-06-12
GET external_subscription by external_id

paulorbpinho-fullstacklabs:
Added new method to get external_subscription by external_id.


amandamfielding:
DevQA +1

PR #487, opened 2025-01-14
fix setting recurly error when using RecurlyErrors class

BartoszRackowski:
`this.recurlyErrors = (RecurlyErrors) recurlyErrors;` is not working
`this.recurlyErrors = (RecurlyErrors) recurlyErrors;` is equivalent to `this.recurlyErrors = this.recurlyErrors;` which assigns value to itself (non initialized, null effectively).

This change is to fix this issue by assigning `recurlyError` from param, also covered by tests.


DamianKaczmarczykK (approved):
Test if I can give approval
EDIT: security is too good to bypass (mission failed)

